PR #11, opened 2023-10-30
Infra setup for recovery and export frames

r-n-o:
This PR has the basics:
* a new Dockerfile which installs and starts an nginx server, serving recovery and export frames on 2 different ports (8080, 8081)
* a new workflow to publish docker images to GHCR
* a new `kustomize` folder with configuration to define a service

## Testing
* I have tested the Dockerfile locally, it works. 
* I have published an image to ghcr
* I've also pulled that same image and ran it in a local `k3d` cluster

Next step: run this in our preprod cluster!


drGrove (approved):

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM docker.io/debian@sha256:741bae561f5c2261f4cdd535e4fd4c248dec0aafc1b9a1410b3d67ad24571340
```
drGrove, line 1: I'd use https://hub.docker.com/r/nginxinc/nginx-unprivileged instead.

nginx.conf
```diff
@@ -0,0 +1,36 @@
...
+    worker_connections 1024;
+}
+
+http {
+    access_log /dev/stdout;
+
+    server {
+       listen 8080;
+       root /home/nginx/recovery;
+       location = /health {
```
drGrove, line 20: Why a separate health endpoint? If you get a 200 from the recovery endpoint isn't it the same thing?

nginx.conf
```diff
@@ -0,0 +1,36 @@
...
+           access_log off;
+           add_header 'Content-Type' 'application/json';
+           return 200 '{"status":"UP"}';
+       }
+    }
+  
+    server {
+        listen 8081;
+        root /home/nginx/export;
+        location = /health {
```
drGrove, line 30: Same as above

Dockerfile
drGrove: If you do go down that road, you should be able to drop the nginx.conf completely

drGrove (dismissed):

nginx.conf
r-n-o: yeah that's a good point. I'll remove it.

Dockerfile
r-n-o: I was trying to go down that road but got stuck on a dumb problem: which architecture should I pick? I tried to determine which architecture we have picked for debian (`741bae561f5c2261f4cdd535e4fd4c248dec0aafc1b9a1410b3d67ad24571340`) and got stuck there for way too long 😮‍💨

r-n-o: I don't think I'll be able to drop the nginx.conf completely though, because I need two ports targeting two different directories (default is 1 folder served on a single port)

james-callahan: I call this script `docker-digest`, it's what I use to get the correct checksum for images:
```
#!/bin/sh

set -euf

image="${1:?Missing image name}"

for image in "$@"; do
	echo >&2 "Fetching checksum for '$image'..."

	MANIFEST="$(skopeo inspect docker://"$image" --raw)"
	echo -n "$MANIFEST" | skopeo manifest-digest /dev/stdin
done
```

nginx.conf
james-callahan: You should *not* do this, the body of health checks is reported in kubernetes logs.
Try and have an endpoint that returns nothing other than e.g. `OK`

r-n-o: Ah. Will add this back then!

Dockerfile
```diff
@@ -0,0 +1,16 @@
+# This is nginx 1.24.0 on bullseye.
+# https://hub.docker.com/layers/nginxinc/nginx-unprivileged/1.24.0-bullseye/images/sha256-a8ec652916ce1e7ab2ab624fe59bb8dfc16a018fd489c6fb979fe35c5dd3ec50
+FROM docker.io/nginxinc/nginx-unprivileged@sha256:a8ec652916ce1e7ab2ab624fe59bb8dfc16a018fd489c6fb979fe35c5dd3ec50
```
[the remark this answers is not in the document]
james-callahan: FYI the correct digest (at least what I get) is:
```
$ docker-digest docker.io/nginxinc/nginx-unprivileged:1.24.0-bullseye
Fetching checksum for 'docker.io/nginxinc/nginx-unprivileged:1.24.0-bullseye'...
sha256:ac0654a834233f7cc95b3a61550a07636299ce9020b5a11a04890b77b6917dc4
```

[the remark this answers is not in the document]
r-n-o: I can repro on my side as well. `sha256:ac0654a834233f7cc95b3a61550a07636299ce9020b5a11a04890b77b6917dc4` is the correct digest 👍

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# https://hub.docker.com/layers/nginxinc/nginx-unprivileged/1.24.0-bullseye/images/sha256-a8ec652916ce1e7ab2ab624fe59bb8dfc16a018fd489c6fb979fe35c5dd3ec50
+FROM docker.io/nginxinc/nginx-unprivileged:1.24.0-bullseye@sha256:ac0654a834233f7cc95b3a61550a07636299ce9020b5a11a04890b77b6917dc4
+
+LABEL org.opencontainers.image.title frames
+LABEL org.opencontainers.image.source https://github.com/tkhq/frames
+
+COPY nginx.conf /etc/nginx/nginx.conf
+
+COPY recovery /usr/share/nginx/recovery
+COPY export /usr/share/nginx/export
```
james-callahan, line 11: I don't think these `COPY` statements will be reproducible (i.e. the timestamps of the directories will change each time this runs)

I'm not 100% sure we need to address that now, but I think the fix would be something like:
1. COPY into a `/newroot` 
2. fix timestamps in `/newroot`
3. Do a `FROM scratch` + `COPY --from` to flatten the COPYs away

james-callahan (approved):

Dockerfile
r-n-o: Yeah it's probably something we'll want to address in a followup, either with this trick or by getting toolchain involved: tracking this here: https://github.com/tkhq/frames/issues/13

PR #17, opened 2023-11-16
Send `ERROR` event when something goes wrong in event processing logic

r-n-o:
Pretty simple change: let's try/catch around events we're receiving / processing and "throw" errors via `postMessage`. Parent pages do not have the ability to catch errors within cross-domain iframes unfortunately.

As step (2) I'll listen for these events in our SDK and make sure to bubble them as errors by `reject`ing the promise so callers can chain (`injectBundle.then(...).catch(...)`)

### DIFF
--- a/export/index.html
+++ b/export/index.html
@@ -340,18 +340,30 @@
 
       // TODO: find a way to filter messages and ensure they're coming from the parent window?
       // We do not want to arbitrarily receive messages from all origins.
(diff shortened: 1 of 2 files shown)


oliviathet (approved):
ah, much-needed fix and plan for step 2!

emostov (approved):

PR #18, opened 2023-11-20
Add base58check decoding to recovery iframe

r-n-o:
This branch adds base58check decoding. For now I've left base64url decoding in place, but we should be able to get rid of it once we release base58check recovery codes.

The nice thing about recovery codes is that they auto-expire after 30mins. So as soon as we switch to base58check codes, clock starts ticking, and we're good to delete the awkward branching after just 30mins (although, practically we'll wait a few hours at least!)


.nvmrc
emostov, line 1: 👏

recovery/index.html
```diff
@@ -262,6 +262,64 @@ <h2>Message log</h2>
...
+        var decodedBytes = [];
+        var result = [];
+        for (var i = 0; i < s.length; i++) {
+          if (alphabet.indexOf(s[i]) === -1) {
+            throw new Error(`cannot base58-decode: ${s[i]} isn't a valid character`)
+          }
+          var carry = alphabet.indexOf(s[i]);
+
+          // If the current base58 digit is 0, append a 0 byte.
+          // "result.length == i" can only happen if we have not seen non-zero bytes so far
```
emostov, line 289: don't follow this second line. Reads to me like `result.length == i` will only be true if you pushed non 0-bits. But I think you mean it means that you have seen at least 1 byte

recovery/index.html
```diff
@@ -262,6 +262,64 @@ <h2>Message log</h2>
...
+          if (carry == 0 && result.length === i) {
+            result.push(0);
+          }
+
+          var j = 0;
+          while (j < decodedBytes.length || carry > 0) {
+            var currentByte = decodedBytes[j];
+            // shift the current byte 58 units and add the carry amount (or just add the carry amount if this is a new byte)
+            currentByte = currentByte ? currentByte * 58 + carry : carry;
```
emostov, line 298: when would `currentByte` be falsey?

recovery/index.html
r-n-o: `result` actually contains _only_ zero bytes at this stage. So, if we had a non-zero bytes that we iterated over, result is untouched which means its length doesn't change. But `i` keeps increasing. I can change the check to result `i > result.length`. And maybe rename result to `leadingZeros`?

recovery/index.html
```diff
@@ -854,7 +913,26 @@ <h2>Message log</h2>
...
+          && bundle.indexOf("O") === -1 && bundle.indexOf("I") === -1 && bundle.indexOf("l") === -1 && bundle.indexOf("0") === -1
+        ) {
+          // If none of these characters are in the bundle we assume it's a base58check-encoded string
+          // This isn't perfect: there's a small chance that a base64url-encoded string doesn't have any of these characters by chance!
+          // But we accept this risk given this branching is only here to support our transition to base58check.
+          // I hear you'd like to quantify this risk? Let's do it.
+          // Assuming random bytes in our bundle and a bundle length of 33 (public key, compressed) + 48 (encrypted cred) = 81 bytes.
+          // The odds of a byte being in the overlap set between base58 and base64url is 58/64=0.90625.
+          // Which means the odds of a 81 bytes string being in the overlap character set for its entire length is...
+          // ... 0.90625^81 = 0.0003444209703
```
emostov, line 929: Nice

emostov (approved):
Didn't completely wrap my brain around ss58 impl, but the test coverage looks good so I trust it!

recovery/index.html
r-n-o: When its value is `undefined` (`decodedBytes[j]` can be `undefined`):
```
> bytes = [] 
> bytes[0]
undefined
```

r-n-o: I've reworked the code to make this clearer.

recovery/index.html
r-n-o: Ok done in the latest diff, let me know what you think!

emostov:
Readability updates look great 👍

oliviathet (approved):
🚀

PR #19, opened 2023-11-20
add support for auth (via email)

andrewkmin:
$title

Related: https://github.com/tkhq/sdk/pull/169

This will require dns updates to have this hosted at a Turnkey subdomain.


r-n-o:
@andrewkmin do you think we can consolidate recovery and auth under the same iframe? Conceptually they do the same thing?
* spin up a key pair on first load
* accept a bundle (we can have different events for different bundle types, if these are different -- but I think in this case both recovery and auth bundles are the same, they contain an encrypted P256 private key)
* sign / stamp payloads

Are there specific aspects of auth that require a separate iframe? Otherwise we can have a common domain for both.

andrewkmin:
> @andrewkmin do you think we can consolidate recovery and auth under the same iframe? Conceptually they do the same thing?
> 
> * spin up a key pair on first load
> * accept a bundle (we can have different events for different bundle types, if these are different -- but I think in this case both recovery and auth bundles are the same, they contain an encrypted P256 private key)
> * sign / stamp payloads
> 
> Are there specific aspects of auth that require a separate iframe? Otherwise we can have a common domain for both.

This absolutely makes sense. Arguably the trickiest part in all this is naming 😄 between the two, I feel like both fall under the `auth` umbrella more so than the `recover` umbrella, but I'm open to entries here. `injector` would be too general; `secure entry`... hmmm

Dockerfile
```diff
@@ -7,7 +7,7 @@ LABEL org.opencontainers.image.source https://github.com/tkhq/frames
 
 COPY nginx.conf /etc/nginx/nginx.conf
 
-COPY recovery /usr/share/nginx/recovery
+COPY auth /usr/share/nginx/auth
 COPY export /usr/share/nginx/export
```
r-n-o, line 11: I think we need to keep supporting recovery.turnkey.com. So let's "cheat":
```
COPY auth /usr/share/nginx/recovery
```
:)

README.md
```diff
@@ -1,17 +1,22 @@
 # Frames
 
-This repository contains code for the recovery and export components of Turnkey. These components can be embedded as iframes by users to support end-users in recovery and export.
+This repository contains code for the recovery, export, and auth components of Turnkey. These components can be embedded as iframes by users to support end-users.
 
 ## Email Recovery
```
r-n-o, line 5: Update the title to "Auth and Recovery"?

r-n-o:
> Arguably the trickiest part in all this is naming 😄 between the two, I feel like both fall under the auth umbrella more so than the recover umbrella, but I'm open to entries here

Yeah I'm with you there! I'd say `recovery` falls under the `auth` umbrella 👍

auth/index.html
```diff
@@ -863,7 +864,7 @@ <h2>Message log</h2>
       // TODO: find a way to filter messages and ensure they're coming from the parent window?
       // We do not want to arbitrarily receive messages from all origins.
       window.addEventListener("message", async function(event) {
-        if (event.data && event.data["type"] == "INJECT_RECOVERY_BUNDLE") {
+        if (event.data && (event.data["type"] == "INJECT_CREDENTIAL_BUNDLE" || event.data["type"] == "INJECT_RECOVERY_BUNDLE")) {
```
r-n-o, line 867: 👏

auth/favicon.svg
r-n-o, line 1: We might need a different color than usual! Red? Purple?

Doesn't really matter in the end, not many people will visit these iframe as a standalone page.

.github/workflows/deploy.yml
```diff
@@ -14,13 +14,13 @@ jobs:
...
+      - name: Publish Auth
         uses: cloudflare/pages-action@f0a1cd58cd66095dee69bfa18fa5efd1dde93bca #v1.5.0
         with:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
-          projectName: recovery
-          directory: recovery
+          projectName: auth
+          directory: auth
           gitHubToken: ${{ secrets.GITHUB_TOKEN }}
```
r-n-o, line 24: Note: this will fail on the cloudflare side, I'll need to edit the project to be named "auth".

r-n-o (approved):

oliviathet (approved):
wooooh 🎉  this looks great. love seeing `recovery` -> `auth`.

PR #25, opened 2024-02-26
Add import.turnkey.com

oliviathet:
* Add HpkeEncrypt:
  * sets up sender context using the receiver public key (the target public key returned from the enclave in InitImportWallet/InitImport)
  * build aad
  * encrypt plaintext using receiver public key
  * format hex-encoded ciphertext and hex-encoded encapped public key into encrypted bundle
* Add 2 events that post messages up to the parent of the import.turnkey.com iframe:
  * `INJECT_IMPORT_BUNDLE`: injects the `import_bundle` which takes the `target_public_key` from `InitImportWallet`/`InitImportPrivateKey`, imports it as a raw key, exports it as a JWK, then stores it in local storage.
  * `EXTRACT_WALLET_ENCRYPTED_BUNDLE`:  gets whatever text is in the `wallet-bundle` textarea, encrypt it, post message to parent
*  Add standalone version (import.turnkey.com/standalone) for testing/demo purposes
  
  
 Linear ticket: ENG-1079

 TODO (in follow up PR)
 * add basic validation for the wallet seed: trim whitespace and check that it's not empty, check that word count is valid per BIP-39
 * add a new event to listen to that lets parent set height and width of the textarea


import/index.html
```diff
@@ -0,0 +1,298 @@
...
+        {
+          plaintextBuf,
+          receiverPubJwk: targetPublicKeyJwk,
+        });
+
+      // Send up ENCRYPTED_BUNDLE_EXTRACTED message
+      TKHQ.sendMessageUp("ENCRYPTED_BUNDLE_EXTRACTED", encryptedBundle)
+    }
+
+    const HpkeEncrypt = async ({ plaintextBuf, receiverPubJwk }) => {
```
oliviathet, line 257: @emostov does this look right for encryption?

import/index.html
```diff
@@ -0,0 +1,298 @@
...
+    Script defining important helpers.
+    These helpers are unit-testable, so most of the logic should be written here.
+  -->
+  <script>
+    window.TKHQ = function() {
+      /** constants for LocalStorage */
+      const TURNKEY_TARGET_EMBEDDED_KEY = "TURNKEY_TARGET_EMBEDDED_KEY"
+      /** 1 week in milliseconds */
+      const TURNKEY_TARGET_EMBEDDED_KEY_TTL_IN_MILLIS = 1000 * 60 * 60 * 24 * 7;
+
```
r-n-o, line 48: Do we need an embedded key in the case of import? I think it's fine if we generate a brand new key every time: basically whichever key encrypted the payload will be in the import payload. As long as this is the case, this key can be forgotten and the iframe can completely disappear after encrypting the payload.

(and if the user refreshes the page, a new P256 key is created, that shouldn't be a problem)

EDIT: OH WAIT. I'm tired. We do need this because the import bundle will be encrypted to this embedded key right?

import/index.html
```diff
@@ -0,0 +1,298 @@
...
+          } catch (e) {
+            TKHQ.sendMessageUp("ERROR", e.toString());
+          }
+        }
+        if (event.data && event.data["type"] == "EXTRACT_WALLET_ENCRYPTED_BUNDLE") {
+          try {
+            await onExtractWalletEncryptedBundle()
+          } catch (e) {
+            TKHQ.sendMessageUp("ERROR", e.toString());
+          }
```
r-n-o, line 203: I wonder if we should delete this interface and have the iframe automatically send an event "WALLET_ENCRYPTED" with the encrypted content whenever encryption succeeds? That way the parent page injects the import bundle, and then waits for this WALLET_ENCRYPTED event.

Basically encryption should happen when the user presses enter or submits the form that's within the iframe, and the parent page doesn't have to know when this happens. But I don't know if that's a reasonable thing: the parent page may want to have its own "submit" button 🤷

import/index.html
```diff
@@ -0,0 +1,298 @@
...
+        TKHQ.setTargetEmbeddedKey(targetPublicKeyJwk);
+
+        // todo(olivia): verify the signature with the enclave quorum public key once returned in server messages
+        // const targetSignatureBuf = TKHQ.uint8arrayFromHexString(bundleObj.targetPublicSignature);
+        // const quorumPublicBuf = TKHQ.uint8arrayFromHexString(bundleObj.enclaveQuorumPublic);
+        // const quorumKey = TKHQ.importTargetKey(new Uint8Array(quorumPublicBuf));
+        // const verified = await crypto.subtle.verify(quorumKey, targetSignatureBuf, targetPublicBuf);
+        // if (verified === false) {
+        //   throw new Error("verification failed");
+        // }
```
r-n-o, line 229: 💯

import/standalone.html
```diff
@@ -0,0 +1,370 @@
+<!doctype html>
```
r-n-o, line 1: What's the difference between standalone.html and index.html? Possible to consolidate in a single version or are there fundamental differences? I think there are ways to toggle CSS / content based on whether the current page is embedded as an iframe vs not (you can use `if (window != window.top) { /* I'm an iframe! */ }`)

r-n-o (commented):
Great job cleaning up the extraneous code and keeping the import frame code to a minimum!

import/index.html
oliviathet: Yes! This is the public key generated in the enclave and returned in the "import_bundle" returned by init_import_wallet. It's then used to encrypt the seed phrase sent in "encrypted_bundle" in import_wallet.

import/standalone.html
oliviathet: interesting. i think that this logic (toggling content/css based on whether it's embedded in an iframe) would be useful to go back and implement in export.turnkey.com actually! i think for import however, having a separate standalone page for testing might make it easier (ie less error prone) when building the actual page that gets embedded in an iframe. for the page that gets embedded in an iframe, it's currently somewhat simple (just one text box) but i'll be adding more components in there to support styling and validation for different lengths of mnemonics and different formats of private keys. i think given that additional complexity w content/css, i would just prefer to have this testing-purposes standalone page as its own thing.

sorry for this long stream of consciousness! lmk if it's more of a ramble than a clarification

r-n-o: ah, I see. Standalone is for testing purposes only, makes sense to me!

The reason why it'd be good to have a single page is just consistency: if there are problems with scripts on index.html you might not catch them if you're testing with standalone.html (and vice versa!)

import/index.html
oliviathet: oh i like this idea! this flow with a button inside the iframe potentially makes it 2 clicks for the user (click button in iframe then passkey tap for the api call made outside of the iframe)... so wdyt about supporting this as a "styling/settings" option in my follow up or for settings (basically an event that this page listens to for setting which content shows up and what styling to apply to it via a json)

import/standalone.html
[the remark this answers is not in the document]
oliviathet: good idea! added comments to index.html and standalone.html

import/index.html
```diff
@@ -0,0 +1,298 @@
...
+  <!--
+    Script defining important helpers.
+    These helpers are unit-testable, so most of the logic should be written here.
+  -->
+  <script>
+    window.TKHQ = function() {
+      /** constants for LocalStorage */
+      const TURNKEY_TARGET_EMBEDDED_KEY = "TURNKEY_TARGET_EMBEDDED_KEY"
+      /** 1 week in milliseconds */
+      const TURNKEY_TARGET_EMBEDDED_KEY_TTL_IN_MILLIS = 1000 * 60 * 60 * 24 * 7;
```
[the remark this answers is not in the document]
oliviathet: oops i actually made it really long at first because we don't really need it to expire, but meant to eventually remove the expiration entirely. the reason is that this `target_embedded_key` is the public key of the encryption key we're storing on the org. and we're not setting any expiration there.

PR #26, opened 2024-03-12
utf8-encode the exported private key

oliviathet:
Cannot be released until v2024.3.13 release.

Exported and imported private keys will be utf-8 encoded on the clients (iframes like import.turnkey.com and export.turnkey.com as well as the Turnkey CLI) and will only be base58 or hex-encoded in the Turnkey signer enclave.

### DIFF
--- a/export/index.html
+++ b/export/index.html
@@ -272,8 +272,8 @@
        * @param {Uint8Array} privateKeyBytes
        */
       const parseKey = privateKeyBytes => {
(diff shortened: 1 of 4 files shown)


oliviathet:
Closing in favor of https://github.com/tkhq/frames/pull/27. This PR reflects a direction where the key formatting is applied in the enclave, but we will be applying it in the iframe/cli instead.

PR #27, opened 2024-03-13
Apply key formatting

oliviathet:
Add optional `keyFormat` and `publicKey` parameters to `injectKeyExportBundle`. Add `extractKeyEncryptedBundle`.

### DIFF
--- a/export/index.html
+++ b/export/index.html
@@ -380,11 +380,11 @@
               throw new Error("public key must be specified for SOLANA key format");
             }
             if (privateKeyBytes.length !== 32) {
-              throw new Error(`invalid private key length. Expected 32 bytes. Got ${privateKeyBytes.length()}.`);
+              throw new Error(`invalid private key length. Expected 32 bytes. Got ${privateKeyBytes.length}.`);
             }
             const publicKeyBytes = uint8arrayFromHexString(publicKey);
             if (publicKeyBytes.length !== 32) {
-              throw new Error(`invalid public key length. Expected 32 bytes. Got ${privateKeyBytes.length()}.`);
+              throw new Error(`invalid public key length. Expected 32 bytes. Got ${publicKeyBytes.length}.`);
             }
             const concatenatedBytes = new Uint8Array(64);
             concatenatedBytes.set(privateKeyBytes, 0);
--- a/export/index.html
+++ b/export/index.html
@@ -272,8 +272,8 @@
        * @param {Uint8Array} privateKeyBytes
        */
       const parseKey = privateKeyBytes => {
-        const privateKeyHexString = uint8arrayToHexString(privateKeyBytes);
-        return "0x" + privateKeyHexString;
+        const decoder = new TextDecoder("utf-8");
+        return decoder.decode(privateKeyBytes);
       }
 
       /**
--- a/export/index.html
+++ b/export/index.html
@@ -268,7 +268,7 @@
       }
 
       /**
-       * Returns a hex-encoded raw private key from private key bytes.
+       * Returns a hex-encoded or base58-encoded private key from private key bytes.
        * @param {Uint8Array} privateKeyBytes
        */
       const parseKey = privateKeyBytes => {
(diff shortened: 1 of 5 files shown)


export/index.html
```diff
@@ -268,12 +282,120 @@ <h2>Message log</h2>
...
+       * the encoding and format specified by `keyFormat`. Defaults to
+       * hex-encoding if `keyFormat` isn't passed.
        * @param {Uint8Array} privateKeyBytes
+       * @param {string} keyFormat Can be "HEXADECIMAL" or "SOLANA"
+       * @param {string} publicKey Hex-encoded public key, needed for Solana private keys
        */
-      const parseKey = privateKeyBytes => {
-        const privateKeyHexString = uint8arrayToHexString(privateKeyBytes);
-        return "0x" + privateKeyHexString;
+      const parseKey = (privateKeyBytes, keyFormat, publicKey) => {
```
r-n-o, line 376: This isn't a parsing (text => bytes) function, it's the other way around! We go from bytes to text. So I think this should be "encodeKey" or "encodePrivateKey"?

export/index.html
```diff
@@ -377,6 +501,8 @@ <h2>Message log</h2>
         window.postMessage({
           "type": "INJECT_KEY_EXPORT_BUNDLE",
           "value": document.getElementById("key-export-bundle").value,
+          "keyFormat": document.getElementById("key-export-format").value,
+          "publicKey": document.getElementById("key-export-public-key").value,
```
r-n-o, line 505: We could potentially derive the public key from the private key in the iframe to avoid this extra param, but this means supporting ed25519. Not sure it's worth the complexity (webcrypto doesn't support it natively)

timurnkey: how do they end up getting the public key? is it sent down in the bundle?

import/standalone.html
```diff
@@ -140,6 +153,84 @@ <h2>Message log</h2>
...
+      }
+
+      /**
+       * Returns private key bytes from a private key, represented in
+       * the encoding and format specified by `keyFormat`. Defaults to
+       * hex-encoding if `keyFormat` isn't passed.
+       * @param {string} privateKey
+       * @param {string} keyFormat Can be "HEXADECIMAL" or "SOLANA"
+       */
+       const formatKey = (privateKey, keyFormat) => {
```
timurnkey, line 212: nit: would name this the opposite of the other function:`decodeKey`

export/index.html
oliviathet: yep, currently it'd get sent down from the js sdk so they would need to fetch the private key's public key when the user exports it... i can look into supporting the curve because it would be nice to only ask for the key format field. having two optional fields feels like extra education for devs.

oliviathet: @r-n-o @timurnkey i haven't had time to look into this today, but wdyt about adding https://www.npmjs.com/package/@noble/ed25519 as an ES module just to derive the public key for a private key when `solana` is passed as the key format? not ideal to be using a CDN but it could be a short term solution until we can implement this ourselves (and/or when webcrypto supports it) that allows for a long term solution of not needing to pass the public key param

r-n-o: Yeah we could pull a dependency for it for sure. We could also grab just the code we need and inline it with attribution. At the end of the day what we need is "just" a bigInt multiplication / exponentiation to go from private key (int) to public key (point on the curve).

timurnkey: +1 to inlining since it's a sensitive flow.

Would this dependency be an attacker vector? idk 🤷 Are there ways a ES module dependency could eject data out of the iframe (ajax, events, extensions)?

oliviathet: I agree that inlining + attribution is the ideal solution. Since we need to get this fix out tomorrow (this would be our third delay otherwise), wdyt about shipping a temporary fix (https://github.com/tkhq/frames/pull/28) that uses a ES module dependency and then Monday I can start working on the inline replacement.

r-n-o (approved):

PR #30, opened 2024-03-27
Harden `uint8arrayFromHexString`

r-n-o:
Pretty straightforward: this change adds some validation to the argument passed to `uint8arrayFromHexString` and errors explicitly if it's bad (not a string, or an empty string, or a non-hex string)

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -213,7 +213,8 @@
        * @returns {Uint8Array}
        */
       var uint8arrayFromHexString = function(hexString) {
-        if (!hexString || hexString.length % 2 != 0 || !/^[0-9A-Fa-f]+$/.test(hexString)) {
+        var hexRegex = /^[0-9A-Fa-f]+$/;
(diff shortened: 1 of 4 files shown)


auth/index.html
```diff
@@ -213,6 +213,9 @@ <h2>Message log</h2>
        * @returns {Uint8Array}
        */
       var uint8arrayFromHexString = function(hexString) {
+        if (!hexString || hexString.length % 2 != 0 || !/^[0-9A-Fa-f]+$/.test(hexString)) {
```
timurnkey, line 216: nit: we could move the regex into a variable so it's slightly more readable

```
var hexRegex = /^[0-9A-Fa-f]+$/;
if (!hexString || hexString.length % 2 != 0 || !hexRegex.test(hexString)) {
...
}
```

r-n-o: missed opportunity to name this `regHex` 😅

timurnkey (approved):

PR #31, opened 2024-04-02
ENG-1446: Verify enclave sig in export/import

oliviathet:
Based on ToB findings. Pin the signer enclave's quorum public key to import and export sites and verify the signed public keys from the enclave. Once we've deployed a new release that returns the enclave quorum public key in the import and export bundles, we can implement the TODOs I've left for checking that those fields aren't null and can be compared to the pinned quorum public key -- this lets the client know when the pinned quorum public key should be rotated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        directory: ["auth", "export"]
+        directory: ["auth", "export", "import"]
 
     steps:
       - name: Checkout
--- a/export/index.html
+++ b/export/index.html
@@ -124,6 +124,28 @@
       /** Turnkey Signer enclave's public key */
       const TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY = "04ca7c0d624c75de6f34af342e87a21e0d8c83efd1bd5b5da0c0177c147f744fba6f01f9f37356f9c617659aafa55f6e0af8d169a8f054d153ab3201901fb63ecb04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569"
 
+      /*
+      * Import a key to encrypt to as a CryptoKey and export it as a JSON Web Key.
+      */
+      const importTargetKey = async targetPublic => {
+        const targetKey = await crypto.subtle.importKey("raw", targetPublic, {
+          name: 'ECDH',
+          namedCurve: 'P-256',
+        }, true, []);
+
+        return await crypto.subtle.exportKey("jwk", targetKey);
+      }
+
+      /*
+      * Imports the quorum public key as a CryptoKey and returns it.
+      */
+      const importQuorumKey = async quorumPublic => {
+        return await crypto.subtle.importKey("raw", quorumPublic, {
+          name: 'ECDSA',
+          namedCurve: 'P-256'
+        }, true, ['verify']);
+      }
+
       /**
        * Creates a new public/private key pair and persists it in localStorage
        */
@@ -190,7 +212,7 @@
        * expiry time is missing.
        * @param {string} key
        */
-      const getItemWithExpiry = (key) => {
+      const getItemWithExpiry = key => {
         const itemStr = window.localStorage.getItem(key);
         if (!itemStr) {
           return null;
@@ -214,7 +236,7 @@
        * @param {string} hexString
        * @returns {Uint8Array}
        */
-       const uint8arrayFromHexString = function(hexString) {
+       const uint8arrayFromHexString = hexString => {
         var hexRegex = /^[0-9A-Fa-f]+$/;
         if (!hexString || hexString.length % 2 != 0 || !hexRegex.test(hexString)) {
           throw new Error('cannot create uint8array from invalid hex string: "' + hexString + '"');
@@ -245,7 +267,7 @@
       /**
        * Function to verify enclave signature on import bundle received from the server. 
        */
-       const verifyEnclaveSignature = async (enclaveQuorumPublic, publicKey, publicSignature) => {
+       const verifyEnclaveSignature = async (enclaveQuorumPublic, publicSignature, publicKey) => {
         // Second half is the public key for the enclave quorum encryption key
         const encryptionQuorumPublic = TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY.slice(130, 260);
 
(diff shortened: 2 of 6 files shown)


export/index.html
```diff
@@ -121,6 +121,18 @@ <h2>Message log</h2>
       const TURNKEY_EMBEDDED_KEY = "TURNKEY_EMBEDDED_KEY"
       /** 48 hours in milliseconds */
       const TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS = 1000 * 60 * 60 * 48;
+      /** Turnkey Signer enclave's public key */
+      const TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY = "04ca7c0d624c75de6f34af342e87a21e0d8c83efd1bd5b5da0c0177c147f744fba6f01f9f37356f9c617659aafa55f6e0af8d169a8f054d153ab3201901fb63ecb04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569"
+
+      /*
+      * Imports the quorum public key as a CryptoKey and returns it.
+      */
+      const importQuorumKey = async quorumPublic => {
```
timurnkey, line 130: let's call this `loadX` instead so it's less confusing since these are import/export flows

import/index.html
```diff
@@ -56,6 +58,16 @@
         return await crypto.subtle.exportKey("jwk", targetKey);
       }
 
+      /*
+      * Imports the quorum public key as a CryptoKey and returns it.
+      */
+      const importQuorumKey = async quorumPublic => {
```
timurnkey, line 64: same

.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        directory: ["auth", "export"]
+        directory: ["auth", "export", "import"]
```
r-n-o, line 15: woah. Good catch!

export/index.html
```diff
@@ -240,6 +252,76 @@ <h2>Message log</h2>
         return new Uint8Array([...s, ...r]);
       }
 
+      /**
+       * Function to verify enclave signature on import bundle received from the server. 
+       */
+       const verifyEnclaveSignature = async (enclaveQuorumPublic, publicSignature, publicKey) => {
+        // Second half is the public key for the enclave quorum encryption key
+        const encryptionQuorumPublic = TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY.slice(130, 260);
```
r-n-o, line 260: Wondering if we should commit just what we need to this repo? So basically, instead of:
```
const TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY = "04ca7c0d624c75de6f34af342e87a21e0d8c83efd1bd5b5da0c0177c147f744fba6f01f9f37356f9c617659aafa55f6e0af8d169a8f054d153ab3201901fb63ecb04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569"
```

We'd have:
```
const TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY = "04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569"
```

This will be less surprising.

export/index.html
```diff
@@ -240,6 +252,76 @@ <h2>Message log</h2>
...
+
+        // todo(olivia): throw error if enclave quorum public is null once server changes are deployed
+        if (enclaveQuorumPublic) {
+          if (enclaveQuorumPublic !== encryptionQuorumPublic) {
+            throw new Error("enclave quorum public keys from client and bundle do not match")
```
r-n-o, line 265: Would be good to template the values when they're different. If we get a report of this error in the wild we'll definitely need this information.

export/index.html
```diff
@@ -240,6 +252,76 @@ <h2>Message log</h2>
         return new Uint8Array([...s, ...r]);
       }
 
+      /**
+       * Function to verify enclave signature on import bundle received from the server. 
+       */
+       const verifyEnclaveSignature = async (enclaveQuorumPublic, publicSignature, publicKey) => {
```
timurnkey, line 258: do you know if it's possible to use [JS modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) within an iFrame? I'm seeing a lot of duplicate code which will become harder to maintain. It'd be awesome if we could use a module system although it might not be possible,

export/index.html
```diff
@@ -240,6 +252,76 @@ <h2>Message log</h2>
...
+        const encryptionQuorumPublic = TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY.slice(130, 260);
+
+        // todo(olivia): throw error if enclave quorum public is null once server changes are deployed
+        if (enclaveQuorumPublic) {
+          if (enclaveQuorumPublic !== encryptionQuorumPublic) {
+            throw new Error("enclave quorum public keys from client and bundle do not match")
+          }
+        }
+  
+        // The ECDSA signature is ASN.1 DER encoded but WebCrypto uses raw format
```
r-n-o, line 269: Oh damn 🎖️

r-n-o (approved):
No major concerns here! Small things below but otherwise: 🚀

export/index.html
oliviathet: ya i was going back and forth with this bc one day we might open source the quorum public keys in their concatenated form. but that is def a future problem and a trivial one anyway so i'm down to remove the extra friction that comes with pinning the full key and mysteriously slicing it in half haha (for both here and tkcli which i can update as well)

r-n-o: Yeah it's no big deal either way honestly, but it just invites questions to see a slice on public key bytes 👀

oliviathet: 100% agree haha

export/index.html
oliviathet: yep! the HpkeEncrypt and HpkeDecrypt code are actually already JS modules. at some point early on, i tried to DRY up the shared code across the 3 pages but it is a lot of refactoring. as some follow ups, i think we modularize:
- part 1: import/index.html and import/standalone.html
- part 2: import and export
- part 3: import/export and auth

oliviathet:
@r-n-o Updated the padding function, now called `normalizePadding`. Added some unit tests for `normalizePadding` and `fromDerSignature`. And even though the arrow vs declarative functions weren't the culprit this time, it did trip me up earlier so for now, I just converted all the arrow functions to declarative. In the future when we DRY up some of the code into JS modules, there could be smaller modules of code that make it easier to go back to using arrow functions without encountering a runtime issue

export/index.test.js
```diff
@@ -160,4 +161,78 @@ describe("TKHQ", () => {
     // TODO: test logMessage / sendMessageUp
     expect(true).toBe(true);
   })
+
+  it("normalizes padding in a byte array", () => {
```
r-n-o, line 165: ❤️

r-n-o (approved):

PR #32, opened 2024-04-02
ENG-1443: Reset iframe embedded key

oliviathet:
Based on ToB findings. Reset iframe embedded key in export and import after decryption and encryption respectively


r-n-o (approved):
💯

PR #34, opened 2024-04-11
Apply settings to import

oliviathet:
Allow for settings to be applied to surface-level/UI  components on import.turnkey.com. Currently, the only setting we're supporting is `styles` which is a strictly-defined list of CSS style classes and their corresponding validation/sanitization rules defined in regular expression. Upon receiving the `APPLY_SETTINGS` event from the parent website, import.turnkey.com performs the necessary checks on the settings object. The valid settings object is then applied and persisted in local storage. Upon every page load, the settings value in local storage is retrieved and validated/sanitized again before being applied.

Possible follow-ups: 
* Allow styling to the key div HTML element in export.turnkey.com.
* Allow supported list of mnemonic lengths for import.turnkey.com
* Allow supported list of key formats for import.turnkey.com and export.turnkey.com
* ???


import/index.html
```diff
@@ -292,6 +309,91 @@
...
+      /**
+       * Function to validate and sanitize the styles object using the accepted map of style keys and values (as regular expressions).
+       * Any invalid style throws an error. Returns an object of valid styles.
+       * @param {Object} styles
+       * @return {Object}
+       */
+       function validateStyles(styles, element) {
+        const validStyles = {};
+
+        const cssValidationRegex = {
```
r-n-o, line 321: 👏

import/index.html
```diff
@@ -292,6 +309,91 @@
...
+          "padding": "^(\\d+(px|em|%|rem) ?){1,4}$",
+          "margin": "^(\\d+(px|em|%|rem) ?){1,4}$",
+          "borderWidth": "^(\\d+(px|em|rem) ?){1,4}$",
+          "borderStyle": "^(none|solid|dashed|dotted|double|groove|ridge|inset|outset)$",
+          "borderColor": "^(transparent|inherit|initial|#[0-9a-f]{3,8}|rgba?\\(\\d{1,3}, \\d{1,3}, \\d{1,3}(, \\d?(\\.\\d{1,2})?)?\\)|hsla?\\(\\d{1,3}, \\d{1,3}%, \\d{1,3}%(, \\d?(\\.\\d{1,2})?)?\\))$",
+          "borderRadius": "^(\\d+(px|em|%|rem) ?){1,4}$",
+          "fontSize": "^(\\d+(px|em|rem|%|vh|vw|in|cm|mm|pt|pc|ex|ch|vmin|vmax))$",
+          "fontWeight": "^(normal|bold|bolder|lighter|\\d{3})$",
+          "fontFamily": "^[^\";<>]*$", // checks for the absence of some characters that could lead to CSS/HTML injection
```
r-n-o, line 330: Nit: it'd be nice to have some unit tests asserting that these characters aren't allowed? You already have a harness for tests, let's use it!

r-n-o (approved):
Very nice!!

PR #35, opened 2024-04-19
separate prod and preprod

andrewkmin:


james-callahan:
It seems weird that the one docker container is always serving both.

Could you make it an argument when starting the daemon?

r-n-o (approved):
Getting this in is necessary for us to test export on wallet preprod, and it's holding the current release back. So let's merge as-is and we can make separate nginx.conf files for preprod and prod respectively, and then pass these to the daemon with `nginx -c <file>` via a `cmd` override in gitops.

PR #36, opened 2024-04-23
replace ed25519 loaded from cdn

oliviathet:
ENG-1479

### DIFF
--- a/export/index.html
+++ b/export/index.html
@@ -783,9 +783,7 @@
         const privateKeyBytes = new Uint8Array(keyBytes);
         if (keyFormat === "SOLANA") {
           const privateKeyHex = TKHQ.uint8arrayToHexString(privateKeyBytes.subarray(0,32));
-          console.log(privateKeyHex);
           const publicKeyBytes = TKHQ.getEd25519PublicKey(privateKeyHex);
-          console.log(TKHQ.uint8arrayToHexString(publicKeyBytes.subarray(0,32)));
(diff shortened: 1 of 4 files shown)


export/index.html
```diff
@@ -774,7 +783,7 @@ <h2>Message log</h2>
         const privateKeyBytes = new Uint8Array(keyBytes);
         if (keyFormat === "SOLANA") {
           const privateKeyHex = TKHQ.uint8arrayToHexString(privateKeyBytes.subarray(0,32));
-          const publicKeyBytes = ed.getPublicKey(privateKeyHex);
+          const publicKeyBytes = TKHQ.getEd25519PublicKey(privateKeyHex);
```
andrewkmin, line 786: 👨‍🍳💋

andrewkmin (approved):
👏👏👏

🔥🔥🔥🚒🧨🧑‍🚒🧯

export/index.html
```diff
@@ -565,10 +578,6 @@ <h2>Message log</h2>
   <script type="module">
     // TODO: this should be bundled at build time or replaced with code written by Turnkey entirely.
     import * as hpke from "https://esm.sh/@hpke/core";
```
andrewkmin, line 580: out of scope for this PR but I'd assume we'll want to tackle this as well at some point?

oliviathet: yep! this one's a big one so deserves a separate PR. also given that it's used by auth, recovery, export, and import, might want to reorganize the code a bit while we do it so there's less duplication

PR #40, opened 2024-05-09
Fix export signature verification

r-n-o:
Draft to retrieve the actual signed bytes from the bundle.

### DIFF
--- a/export/index.template.html
+++ b/export/index.template.html
@@ -749,7 +749,12 @@
             throw new Error(`unable to find signed data in bundle: ${bundle}`);
           }
 
-          verified = await TKHQ.verifyEnclaveSignature(bundleObj.enclaveQuorumPublic, bundleObj.dataSignature, signedData + "}");
+          // We add the closing brace back: it was our marker for the end of the group in the regex above, hence not captured.
+          const signedDataBytes = new TextEncoder().encode(signedData[1] + "}");
+          // Encode these bytes to hex since verifyEnclaveSignature expects hex-encoded bytes
+          const signedDataHexString = TKHQ.uint8arrayToHexString(signedDataBytes);
+
+          verified = await TKHQ.verifyEnclaveSignature(bundleObj.enclaveQuorumPublic, bundleObj.dataSignature, signedDataHexString);
           if (!verified) {
             throw new Error(`failed to verify enclave signature: ${bundle}`);
           }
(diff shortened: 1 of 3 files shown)


export/index.template.html
```diff
@@ -742,7 +742,19 @@ <h2>Message log</h2>
...
+          if (signedData === null || signedData.length !== 2) {
+            throw new Error(`unable to find signed data in bundle: ${bundle}`);
+          }
+
+          // We add the closing brace back: it was our marker for the end of the group in the regex above, hence not captured.
+          const signedDataBytes = new TextEncoder().encode(signedData[1] + "}");
+          // Encode these bytes to hex since verifyEnclaveSignature expects hex-encoded bytes
+          const signedDataHexString = TKHQ.uint8arrayToHexString(signedDataBytes);
+
+          verified = await TKHQ.verifyEnclaveSignature(bundleObj.enclaveQuorumPublic, bundleObj.dataSignature, signedDataHexString);
```
andrewkmin, line 757: tiny nit: think it might make sense to rename the `publicKey` argument of `TKHQ.verifyEnclaveSignature` for consistency. but definitely a huge stamp of approval for now 🥳

r-n-o: yeah that's fair. Let's rename for clarity!

import/index.template.html
```diff
@@ -546,7 +549,16 @@
...
             throw new Error("method not loaded");
           }
-          verified = await TKHQ.verifyEnclaveSignature(bundleObj.enclaveQuorumPublic, bundleObj.dataSignature, bundleObj.data);
+
+          // Temporary solution to get the bytes signed by the enclave
+          var signedData = /"data":({[^}]+)/.exec(bundle);
+          if (signedData === null || signedData.length !== 2) {
+            throw new Error(`unable to find signed data in bundle: ${bundle}`);
+          }
+          const signedDataBytes = new TextEncoder().encode(signedData[1] + "}");
```
andrewkmin, line 558: could add the same comments from `export/index.template.html`, but totally optional

andrewkmin (approved):
🏅

r-n-o:
Closing this PR, we've decided to fix this at the enclave level instead. The idea: return the actual signed bytes. That way there's no funny regex/parsing business that needs to happen, we can just grab this from the bundle and call verifySignature without further ado.

PR #41, opened 2024-05-09
Fix import & export bundle signature verification

r-n-o:
Followup to #40 -- but doing it "right". We're changing the format of the bundles to ship bytes so signature verification is a no brainer. The more "complex" part becomes parsing these bytes into JSON but I think the structure of the bundles is now much cleaner and easier to reason about: data (bytes), signature (bytes), public key (bytes).

Example export bundle
```
{"version":"v1.0.0","data":"7b22656e6361707065645075626c6963223a2230346261356136343463393333616465616663646436373164666532616135356466613435316334346666666662343262323064346238316131376561333166363466326164616638383938633163653239666435633565666434373234643862643166396365633031643139396364363030303134666135653538346163353961222c2263697068657274657874223a22343730336265363236346264346432306665653735636634316233326261303265353432623064346264313837306662623233343165363261363634623065343732366531653139313163313234323436386635343430313239613266386564363531666239656232393932303261623231653638633234653462623962363165363364323231353033343339613566343363636561646462613166356262333563356634643762613332386331393464326565656263316433356365313637653261353931222c226f7267616e697a6174696f6e4964223a2238363233356265642d623235392d343464612d396162662d393236373831653037626534227d","dataSignature":"30450221009462106686ba5a844a827c9c5f3f0d1b107710a632663d7a98f96c99cf96298a02206db3ccbd99169dea30f90f24f994c18bc7a78ae902373df60a805e058186a6da","enclaveQuorumPublic":"04bce6666ca6c12e0e00a503a52c301319687dca588165b551d369496bd1189235bd8302ae5e001fde51d1e22baa1d44249f2de9705c63797316fc8b7e3969a665"}
```


andrewkmin (approved):
👨‍🍳👨‍🍳👨‍🍳

PR #46, opened 2024-08-04
Pin hpke version to 1.2.7

moe-dev:


r-n-o (approved):
Tested locally, embedded keys are back!

PR #48, opened 2024-08-06
Pad export iframes jwk

omkarshanbhag:


export/index.template.html
```diff
@@ -399,12 +399,67 @@ <h2>Message log</h2>
...
+        const publicKey =  await window.crypto.subtle.importKey(
+          "jwk", 
+          {
+              kty: jwkPrivateCopy.kty,
+              crv: jwkPrivateCopy.crv,
+              d: padBase64Url(jwkPrivateCopy.d, 32),
+              x: padBase64Url(jwkPrivateCopy.x, 32),
+              y: padBase64Url(jwkPrivateCopy.y, 32),
+              ext: jwkPrivateCopy.ext,
+          },
```
r-n-o, line 411: I'm not sure this is correct: we specifically delete `d` here so that it's not part of the jwk. I also don't think there's a need to pad because we start with a valid JWK (the private key) and simply delete the `d` component to re-export a verifying public key which doesn't have the private information (`d`).

omkarshanbhag: good catch with the d field. Yeah we found testing locally that there wasn't variance in the key length -- discussed with @andrewkmin that it might be good to push anyway.

export/index.template.html
```diff
@@ -228,16 +228,23 @@ <h2>Message log</h2>
...
 
       /**
-       * Takes a hex string (e.g. "e4567ab") and returns an array buffer (Uint8Array)
+       * Takes a hex string (e.g. "e4567abc") and returns an array buffer (Uint8Array)
        * @param {string} hexString
+       * @param {number} length: optional expected length of the resulting buffer
        * @returns {Uint8Array}
        */
-       function uint8arrayFromHexString(hexString) {
+       var uint8arrayFromHexString = function(hexString, length) {
```
andrewkmin, line 236: Doesn't really matter much, but any particular reason we're now declaring the function in this way? (as opposed to just `function uint8...`)

export/index.template.html
```diff
@@ -399,12 +406,66 @@ <h2>Message log</h2>
...
+     * @param {string} base64UrlString
+     * @param {number} length: optional number of bytes contained in the resulting string
+     * @return {string}
+     */
+    var padBase64Url = function(base64UrlString, length) {
+      var stringBuffer = base64urlDecode(base64UrlString);
+      var hexString = uint8arrayToHexString(stringBuffer)
+      // Add an extra 0 to the start of the string to get a valid hex string (even length)
+      // (e.g. 0x0123 instead of 0x123)
+      var hexString = hexString.padStart(Math.ceil(hexString.length/2)*2, 0)
```
andrewkmin, line 440: Tiny nit: I realized `padStart`'s second argument should be a string, so in this case `"0"`. We use this in a bunch of places and don't necessarily do it that way everywhere, but commenting as an FYI more than anything else

export/index.template.html
r-n-o: I don't know that it's necessary actually: by definition the values in the JWK are valid, because they're already in the JWK. There's no need to export them, pad them "just to make sure" they're importable since they're part of the input JWK to start with (remember, the error we're seeing is an error to import buffers or numbers into a JWK format!)

The reason why I'm pushing back is the amount of new code is decently large.

export/index.test.js
```diff
@@ -235,3 +235,8 @@ describe("TKHQ", () => {
     ).rejects.toThrow('cannot create uint8array from invalid hex string');
   })
 });
+
+it("pads jwk keys to correct length", async () => {
+  expect(TKHQ.padBase64Url("wqVkltEnJAh2mJ5k9rAxEifyKv2BrAWscZ6QZDCS2Os", 32)).toBe("wqVkltEnJAh2mJ5k9rAxEifyKv2BrAWscZ6QZDCS2Os")
```
andrewkmin, line 240: We can add a little commentary to each of these test cases: the first does not change at all because its underlying byte representation occupies 32 bytes, and thus doesn't require padding

Whereas the second example requires a good amount of padding (you can essentially write out the mini process we went through to go from base64 --> hex / bytes --> base64)

andrewkmin (approved):
LGTM overall!

export/index.template.html
andrewkmin: Yes yes yes

omkarshanbhag: @r-n-o makes sense! thanks for the catch here -- I'll go ahead and close the PR.

r-n-o: No worries! If you wanted you could keep this PR opened and merge the extra test you have. But no big deal either way!

PR #51, opened 2024-08-27
Apply Settings to export

zkharit:
Add support for adding css to the export iframe

### DIFF
--- a/export/index.template.html
+++ b/export/index.template.html
@@ -253,7 +253,7 @@
          */
         function getSettings() {
           const settings = window.localStorage.getItem(TURNKEY_SETTINGS);
-          return settings ?  JSON.parse(settings) : null;
+          return settings ? JSON.parse(settings) : null;
         }
 
         /**
@@ -261,7 +261,10 @@
          * @param {Object} settings
          */
         function setSettings(settings) {
-          window.localStorage.setItem(TURNKEY_SETTINGS, JSON.stringify(settings));
+          window.localStorage.setItem(
+            TURNKEY_SETTINGS,
+            JSON.stringify(settings)
+          );
         }
 
         /**
@@ -695,7 +698,7 @@
          * @param {Object} styles
          * @return {Object}
          */
-         function validateStyles(styles, element) {
+        function validateStyles(styles, element) {
           const validStyles = {};
 
           const cssValidationRegex = {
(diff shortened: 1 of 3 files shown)


export/index.template.html
```diff
@@ -671,6 +689,111 @@ <h2>Message log</h2>
...
+         * Function to validate and sanitize the styles object using the accepted map of style keys and values (as regular expressions).
+         * Any invalid style throws an error. Returns an object of valid styles.
+         * @param {Object} styles
+         * @return {Object}
+         */
+         function validateStyles(styles, element) {
+          const validStyles = {};
+
+          const cssValidationRegex = {
+            padding: "^(\\d+(px|em|%|rem) ?){1,4}$",
```
andrewkmin, line 702: Might make sense to put these valid, modifiable css properties in a doc somewhere

export/index.template.html
```diff
@@ -671,6 +689,111 @@ <h2>Message log</h2>
...
+          if (settingsObj.styles) {
+            // Valid styles will be applied the "key-div" textarea HTML element.
+            const keyDivTextarea = document.getElementById("key-div");
+            if (!keyDivTextarea) {
+              throw new Error(
+                "no key-div HTML element found to apply settings to."
+              );
+            }
+
+            // Validate, sanitize, and apply the styles to the "key-div" textarea.
```
andrewkmin, line 785: The `key-div` element is just a simple div, which later gets its `innerText` set

export/index.template.html
```diff
@@ -709,6 +836,11 @@ <h2>Message log</h2>
       document.addEventListener(
         "DOMContentLoaded",
         async () => {
+          // If styles are saved in local storage, sanitize and apply them.
```
andrewkmin, line 839: You may have already done this, and nothing indicates the contrary, but would be nice to do a sanity check to make sure these loaded settings can be overridden if the user calls `applySettings` again

andrewkmin (approved):
non blocking comments.

let's run prettier here `npm run prettier:write`

export/index.template.html
zkharit: changed this

export/index.template.html
zkharit: Yes this works as exepected!

export/index.template.html
zkharit: good idea, can add something to the UI customization page

PR #52, opened 2024-08-29
vendor @hpke/core v1.2.7

andrewkmin:
$title

Source maps are currently excluded. We don't have a need for them, but can also vendor those files if necessary.


mavrukin (approved):

zkharit (approved):
LGTM 🚀

PR #56, opened 2024-12-09
implement message channel

turnekybc:


auth/index.html
```diff
@@ -503,19 +508,16 @@ <h2>Message log</h2>
...
-                type: type,
-                value: value,
-              },
-              "*"
-            );
+          if (parentFrameMessageChannelPort) {
+            parentFrameMessageChannelPort.postMessage({
+              type: type,
+              value: value,
+            })
```
r-n-o, line 520: for compatibility we should keep sending messages with `window.top.postMessage` (actually, we should probably fix this to be `window.parent.postMessage`). SDKs in the wild who are currently using Turnkey's iframes will need this to keep running smoothly.

A breaking change isn't an option unfortunately -- the other option at our disposal would be to version our deployment and host authv2.turnkey.com which points at the new code. But given the level of effort I think it's easy enough to have logic which says "if there is a channel to use, use it. Otherwise, talk to the parent page.

As SDKs and webpages update to the newer version of Turnkey SDK more and more iframes will use channels, and eventually (maybe) we can retire the code which uses "window.top.postMessage" entirely! But that will be months from now.

auth/index.html
```diff
@@ -1006,56 +1009,53 @@ <h2>Message log</h2>
...
       var CREDENTIAL_BYTES = null;
 
-      document.addEventListener(
-        "DOMContentLoaded",
-        async function () {
+      // persist the MessageChannel object so we can use it to communicate with the parent window
+      var iframeMessagePort = null;
+
+      window.addEventListener("message", async function (event) {
+        if (event.ports?.[0]) {
```
r-n-o, line 1016: is this normal? It feels like a weird notation.

auth/index.html
turnekybc: I updated this to be `window.parent.postMessage` for backwards compatibility. As an aside, do we have tracking / analytics on which versions of our SDKs are live in the wild?

import/index.template.html
```diff
@@ -543,56 +612,33 @@
           // In the case of import, this public key is not used because the client encrypts messages
           // to the server's public key.
           TKHQ.sendMessageUp("PUBLIC_KEY_READY", "");
-          // TODO: find a way to filter messages and ensure they're coming from the parent window?
-          // We do not want to arbitrarily receive messages from all origins.
+
```
r-n-o, line 615: In other templates you moved this function to _after_ the call to `window.addEventListener`. I think that makes sense to do it here as well for consistency?

auth/index.html
```diff
@@ -1014,88 +1115,37 @@ <h2>Message log</h2>
...
-            "click",
-            async function (e) {
-              e.preventDefault();
-              window.postMessage({ type: "RESET_EMBEDDED_KEY" });
-            },
-            false
-          );
+          addDOMEventListeners();
+
+          TKHQ.sendMessageUp("PUBLIC_KEY_READY", targetPubHex);
```
r-n-o, line 1127: I'm worried that if this `sendMessageUp` function is called _before_ `TURNKEY_INIT_MESSAGE_CHANNEL` is received (and thus, the channel set up), then the message will fail to be received on the other end (parent frame) because new SDKs won't listen for raw messages coming from postMessage anymore.

So it seems there's a bit of a race condition, and we should wait to receive "TURNKEY_INIT_MESSAGE_CHANNEL" before broadcasting "PUBLIC_KEY_READY"?

Another way to handle this is to broadcast `PUBLIC_KEY_READY` twice:
* leave this here the way you have it (for old SDKs)
* add a call to `sendMessageUp("PUBLIC_KEY_READY", targetPubHex)` again right after `TKHQ.setParentFrameMessageChannelPort` is called (so that new SDKs receive it as well)

Now that I'm writing this I think that's probably the cleanest option?

import/index.template.html
turnekybc: Yes, agreed. I'll move that for consistency.

turnekybc: Updated.

r-n-o (approved):

PR #64, opened 2025-03-11
add getEmbeddedKey

andrewkmin:
$title

SDK counterpart: https://github.com/tkhq/sdk/pull/533

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -164,7 +164,9 @@
         /**
          * Creates a new public/private key pair and persists it in localStorage
          */
-        var initEmbeddedKey = async function (ttl = TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS) {
+        var initEmbeddedKey = async function (
+          ttl = TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS
+        ) {
           var retrievedKey = await getEmbeddedKey();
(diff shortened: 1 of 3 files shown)


auth/index.html
```diff
@@ -1239,6 +1247,16 @@ <h2>Message log</h2>
...
 
+      /**
+       * Function triggered when GET_EMBEDDED_KEY event is received.
+       * @param {string} requestId: idempotency key
+       */
+      var onGetEmbeddedKey = async function (requestId) {
+        const embeddedKey = await TKHQ.getEmbeddedKey();
+
+        TKHQ.sendMessageUp("EMBEDDED_KEY", embeddedKey, requestId);
+      };
```
r-n-o, line 1258: What's the use case for this? This gives the parent page access to the key that's created in the iframe. Parent apps using this would be better served using the parent's local storage since it'd be equivalent security-wise?

The problem I see is that this change applies equally to parent pages who _want_ to use this event, and parent pages who do _not_. Because this API exists it means we can't guarantee it won't be used.

r-n-o: This is maybe acceptable for auth, but seems especially risky for import/export frames. In the export frame, the embedded key decrypts the export bundle so it should be considered especially sensitive.

r-n-o (approved):

PR #65, opened 2025-03-11
Custom ttl for init embedded key

moe-dev:
linked to: https://github.com/tkhq/sdk/pull/535

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -206,7 +206,6 @@
         };
 
         var setEmbeddedKey = function (targetKey, ttl = TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS) {
-          console.log(ttl)
           return setItemWithExpiry(
             TURNKEY_EMBEDDED_KEY,
             JSON.stringify(targetKey),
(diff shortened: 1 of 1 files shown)


auth/index.html
```diff
@@ -205,11 +205,12 @@ <h2>Message log</h2>
           }
         };
 
-        var setEmbeddedKey = function (targetKey) {
+        var setEmbeddedKey = function (targetKey, ttl = TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS) {
+          console.log(ttl)
```
andrewkmin, line 209: let's remove

auth/index.html
```diff
@@ -1163,7 +1164,12 @@ <h2>Message log</h2>
 
             TKHQ.setParentFrameMessageChannelPort(iframeMessagePort);
 
-            await TKHQ.initEmbeddedKey();
+            if (event.data.dangerouslyOverrideIframeKeyTtl) {
+              await TKHQ.initEmbeddedKey(event.data.dangerouslyOverrideIframeKeyTtl);
+            } else {
+                await TKHQ.initEmbeddedKey();
+            }
```
r-n-o, line 1171: Let's do validation to make sure the TTL passed is a number. In JS a lot of stuff is truthy.

I'd also like to simplify TKHQ.initEmbeddedKey so that it always expects an argument (the TTL). And here we can have:
```js
var iframeKeyTtl = TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS
if (event.data.dangerouslyOverrideIframeKeyTtl && ...validation...) {
    iframeKeyTtl = event.data.dangerouslyOverrideIframeKeyTtl;
}

await TKHQ.initEmbeddedKey(iframeKeyTtl);
```

auth/index.html
r-n-o, line 209: doesn't seem to be necessary anymore; guessing this was for debugging!

r-n-o (approved):

PR #66, opened 2025-03-13
Fix race condition for init embedded key

moe-dev:


auth/index.html
```diff
@@ -1123,7 +1124,9 @@ <h2>Message log</h2>
       document.addEventListener(
         "DOMContentLoaded",
         async function () {
+          await new Promise(resolve => setTimeout(resolve, 1));
```
r-n-o, line 1127: this should be removed :)

moe-dev: yessirr

PR #67, opened 2025-04-02
expose functions to refresh an embedded key

andrewkmin:
$title

Accompanying `tkhq/sdk` change: https://github.com/tkhq/sdk/pull/571

https://github.com/user-attachments/assets/ce6339aa-7d5c-476a-8a62-3e7562bd83a8

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -1090,7 +1090,7 @@
           },
           false
         );
-        // also, reset isn't a completely accurate name for this
+        // Reset isn't a completely accurate name for this because this only clears the key, without regenerating a fresh one
         document.getElementById("reset").addEventListener(
           "click",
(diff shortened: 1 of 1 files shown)


auth/index.html
```diff
@@ -1077,6 +1090,7 @@ <h2>Message log</h2>
           },
           false
         );
+        // also, reset isn't a completely accurate name for this
```
r-n-o, line 1093: I understand this comment but others may not? You can complete the thought by saying "because this only clears the key, without regenerating a fresh one"

r-n-o (approved):

auth/index.html
andrewkmin: Good call!

PR #68, opened 2025-04-02
bugfix: return early if embedded key is null

andrewkmin:
$title

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -1274,6 +1274,12 @@
        */
       var onGetEmbeddedPublicKey = async function (requestId) {
         var embeddedKeyJwk = await TKHQ.getEmbeddedKey();
+
+        // If we don't have an embedded key, return an empty string as the public key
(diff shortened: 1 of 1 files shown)


auth/index.html
```diff
@@ -1274,6 +1274,12 @@ <h2>Message log</h2>
        */
       var onGetEmbeddedPublicKey = async function (requestId) {
         var embeddedKeyJwk = await TKHQ.getEmbeddedKey();
+
+        // If we don't have an embedded key, return an empty string as the public key
+        if (!embeddedKeyJwk) {
```
andrewkmin, line 1279: Need to return early if `embeddedKeyJwk` is null; otherwise, the subsequent lines will error out, and no response will be passed back to the client (aka `iframe-stamper`), leading to a hanging/dangling request

r-n-o (approved):

zkharit (approved):

PR #69, opened 2025-04-03
bugfix: reset handler

andrewkmin:
$title

Added test video to https://github.com/tkhq/sdk/pull/571

### DIFF
--- a/auth/index.html
+++ b/auth/index.html
@@ -233,9 +233,8 @@
         /**
          * Removes the embedded key from local storage.
          */
-        var onResetEmbeddedKey = function () {
+        var resetEmbeddedKey = function () {
           window.localStorage.removeItem(TURNKEY_EMBEDDED_KEY);
-          document.getElementById("embedded-key").value = "";
         };
 
         /**
@@ -1027,7 +1026,7 @@
           getItemWithExpiry,
           getEmbeddedKey,
           setEmbeddedKey,
-          onResetEmbeddedKey,
+          resetEmbeddedKey,
           importCredential,
           compressRawPublicKey,
           uncompressRawPublicKey,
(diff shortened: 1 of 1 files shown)


auth/index.html
```diff
@@ -1132,7 +1131,7 @@ <h2>Message log</h2>
         if (event.data && event.data["type"] == "RESET_EMBEDDED_KEY") {
```
andrewkmin, line 1131: Note: I've verified that even older versions of `@tkhq/iframe-stamper` never used `RESET_EMBEDDED_KEY`. We can be confident that changing this handler won't affect any existing clients.

auth/index.html
```diff
@@ -233,9 +233,8 @@ <h2>Message log</h2>
         /**
          * Removes the embedded key from local storage.
          */
-        var onResetEmbeddedKey = function () {
+        var resetEmbeddedKey = function () {
```
andrewkmin, line 236: For readers: functions named `on{action}` differ from those named `{action}` in that the latter are the the main utility functions, and the former are handlers.

So `onResetEmbeddedKey()` gets called when an inbound message is received (i.e. from the parent), and `resetEmbeddedKey()` is responsible for the main guts of the logic (i.e. clearing the embedded key from local storage)

auth/index.html
```diff
@@ -1322,6 +1322,16 @@ <h2>Message log</h2>
...
       };
 
+      /**
+       * Removes the embedded key from local storage.
+       */
+      var onResetEmbeddedKey = async function (requestId) {
+        await TKHQ.resetEmbeddedKey();
+        document.getElementById("embedded-key").value = "";
+
+        TKHQ.sendMessageUp("EMBEDDED_KEY_RESET", null, requestId);
```
andrewkmin, line 1332: Side note: these event names that we pass back up (e.g. `EMBEDDED_KEY_RESET`) aren't really used any more; instead we use `requestId` to match up + handle requests on the parent (iframe client) side

turnekybc (approved):

r-n-o (approved):

PR #70, opened 2025-04-10
oauth proxies

moeodeh3:


Dockerfile
```diff
@@ -2,22 +2,31 @@
 # https://hub.docker.com/layers/nginxinc/nginx-unprivileged/1.24.0-bullseye/images/sha256-a8ec652916ce1e7ab2ab624fe59bb8dfc16a018fd489c6fb979fe35c5dd3ec50
 FROM docker.io/nginxinc/nginx-unprivileged:1.24.0-bullseye@sha256:ac0654a834233f7cc95b3a61550a07636299ce9020b5a11a04890b77b6917dc4
 
-LABEL org.opencontainers.image.title frames
```
zkharit, line 5: this might need to be reverted back to using the `=` and the `""` and same with the following line

Dockerfile
```diff
@@ -2,22 +2,31 @@
...
 
-EXPOSE 8080/tcp
-EXPOSE 8081/tcp
-EXPOSE 8082/tcp
-EXPOSE 8083/tcp
+# oauth
+COPY oauth/origin /usr/share/nginx/origin
+COPY oauth/redirect /usr/share/nginx/redirect
+
+# iframe ports
```
zkharit, line 21: There were previously 4 iframe ports expose, 8080-8083, it looks like there should be one additional port exposed with the addition of the 2 oauth ones

README.md
```diff
@@ -1,19 +1,42 @@
...
 
 This page is hosted at https://auth.turnkey.com/, but we will retain https://recovery.turnkey.com/ for compatibility.
 
-## Key and Wallet Export
+### Key and Wallet Export
 This self-contained HTML page is meant to be used as either a standalone document or to be embedded as an iframe.
 
 This page is hosted at https://export.turnkey.com/
 
+### Key and Wallet Import
```
zkharit, line 19: ⭐️

nginx.conf
```diff
@@ -35,10 +35,12 @@ http {
     # Send access logs to stdout for logging
     access_log /dev/stdout main;
 
-    # Custom server blocks to serve auth and export frames on separate ports.
+     # Custom server blocks to serve auth and export frames on separate ports.
```
zkharit, line 38: nit: spacing

oauth/origin/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "auth-tests",
```
zkharit, line 2: nit: _oauth_-tests?

oauth/origin/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
+  "main": "index.test.js",
+  "scripts": {
+    "start": "serve",
+    "test": "jest",
+    "prettier:check": "prettier --check \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore",
+    "prettier:write": "prettier --write \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore"
+  },
+  "repository": "git@github.com:tkhq/oauth-proxy.git",
```
zkharit, line 11: nit: update repo

oauth/origin/index.html
```diff
@@ -0,0 +1,105 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>Redirecting to Google</title>
```
zkharit, line 5: This will need to be more generic when we support other oauth providers besides just google. I know this as created to solve a google specific problem, but I think theres definitely a world where this provides customers with a realllly easy solution to use oauth with Turnkey

oauth/origin/index.html
```diff
@@ -0,0 +1,105 @@
...
+        if (!provider || !clientId || !redirectUri || !nonce) {
+          displayError(
+            "Error: Missing provider, clientId, redirectUri, or nonce query parameters."
+          );
+          return;
+        }
+
+        const oauthUrl = getOAuthUrl(provider, clientId, redirectUri, nonce);
+        if (!oauthUrl) {
+          // we already displayed an error message so we can just return
```
zkharit, line 96: it might be better to display the error here rather than directly in the `getOAuthUrl` function. A different caller might want to use that function and do its own error handling/display its own more specific error messages based on the result of the function

oauth/origin/index.html
```diff
@@ -0,0 +1,105 @@
...
+  <body>
+    <img class="logo" src="./favicon.svg" alt="Logo" />
+    <h1>Redirecting to Google...</h1>
+
+    <script>
+      (function () {
+        "use strict";
+
+        function displayError(message) {
+          document.body.innerHTML = `<p class="error">${message}</p>`;
```
zkharit, line 52: Not sure if the context is available, but it would be really nice for a user to be able to navigate back to the originating application with a button or something similar, but not a blocker

oauth/redirect/index.html
```diff
@@ -0,0 +1,79 @@
...
+  </head>
+  <body>
+    <h1>Redirecting back to your app...</h1>
+    <p>Please wait...</p>
+    <script>
+      (function () {
+        "use strict";
+
+        const displayError = (message) => {
+          document.body.innerHTML = `<p class="error">${message}</p>`;
```
zkharit, line 44: Same feedback here with the redirect back to originating application button/similar so the user isnt just stuck on this page

oauth/redirect/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "auth-tests",
```
zkharit, line 2: nit: _oauth_-tests

zkharit: maybe oauth-redirect-tests? and the other oauth-origin-tests?

oauth/redirect/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
+  "main": "index.test.js",
+  "scripts": {
+    "start": "serve",
+    "test": "jest",
+    "prettier:check": "prettier --check \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore",
+    "prettier:write": "prettier --write \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore"
+  },
+  "repository": "git@github.com:tkhq/oauth-proxy.git",
```
zkharit, line 11: nit: update repo url

zkharit (commented):
Great addition!!

PR #71, opened 2025-04-11
add origin and redirect

moeodeh3:
https://github.com/user-attachments/assets/8d3a4d07-0b09-413b-8789-56a5ce2c3ca3

https://github.com/user-attachments/assets/499fc3a0-31a5-4622-ada9-8810ced37a4f


origin/index.html
r-n-o, line 1: if we're hosting this at oauth-origin.turnkey.com the folder should be `oauth-origin` for consitency

origin/index.html
```diff
@@ -0,0 +1,152 @@
...
+                let appUrl = scheme;
+                if (!appUrl.includes("://")) {
+                  appUrl += "://";
+                }
+
+                backButton = `<button class="back-button" onclick="window.location.href='${appUrl}'"> < Go back to app</button>`;
+              }
+            }
+          } finally {
+            console.log("I AM HERE");
```
r-n-o, line 84: WHERE ARE YOU?

(let's remove this)

origin/index.html
```diff
@@ -0,0 +1,152 @@
...
+          return `${baseUrl}?${new URLSearchParams(oauthParams).toString()}`;
+        }
+
+        const params = new URLSearchParams(window.location.search);
+
+        // get the provider, clientId, redirectUri, and nonce from the query parameters
+        const provider = params.get("provider");
+        const clientId = params.get("clientId");
+        const redirectUri = params.get("redirectUri");
+        const nonce = params.get("nonce");
```
r-n-o, line 122: is there any sort of validation to perform for these values? Any security implications if we don't?

origin/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  clearMocks: true,
+  setupFilesAfterEnv: ["regenerator-runtime/runtime"],
+  testPathIgnorePatterns: ["/node_modules/"],
+};
```
r-n-o, line 5: I don't see any test files in here. So...probably okay to remove `jest.config.js` as well

redirect/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "redirect-tests",
+  "version": "1.0.0",
```
r-n-o, line 3: Same here; because there is no tests we don't need the boilerplate. Let's remove so there's no confusion about that.

(another option if you feel like it: add a test!)

redirect/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "index.test.js",
+  "scripts": {
+    "start": "serve",
+    "test": "jest",
+    "prettier:check": "prettier --check \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore",
+    "prettier:write": "prettier --write \"**/*.{css,html,js,json,md,ts,tsx,yaml,yml}\" --ignore-path ../.prettierignore"
+  },
+  "repository": "git@github.com:tkhq/oauth-proxy.git",
+  "author": "Turnkey <moe.odeh@turnkey.com>",
```
r-n-o, line 12: unless you want to be spammed personally let's use `hello@turnkey.com`

origin/index.html
moeodeh3: no not really - google enforces it not us 

image

origin/jest.config.js
moeodeh3: was planning on it - but forgot to, done

redirect/package.json
moeodeh3: 😞

nginx.conf
```diff
@@ -50,25 +52,38 @@ http {
...
             try_files /import/$uri /templated/import/$uri /templated/import/$uri/index.html =404;
         }
     }
+
+    # oauth
+    server {
+        listen 8084;
+        root /usr/share/nginx;
+        location / {
+            try_files /oauth-origin/$uri /templated/oauth-origin/$uri /templated/oauth-origin/$uri/index.html =404;
```
r-n-o, line 79: I don't think this is needed because we don't have the need to template anything for these static pages!

(remove `/templated/oauth-origin/$uri /templated/oauth-origin/$uri/index.html`)

nginx.conf
```diff
@@ -50,25 +52,38 @@ http {
...
+        }
+    }
+    server {
+        listen 8085;
+        root /usr/share/nginx;
+        location / {
+            try_files /oauth-redirect/$uri /templated/oauth-redirect/$uri /templated/oauth-redirect/$uri/index.html =404;
```
r-n-o, line 86: likewise here; no templates, no problem

(the reason we have a templating solution for iframes is because we pin the environment's quorum signer public key -- which needs to vary for preprod vs prod)

kustomize/base/resources.yaml
```diff
@@ -20,10 +20,12 @@ spec:
             - sh
             - -c
             - |
-                mkdir -p templated/export templated/import;
+                mkdir -p templated/export templated/import templated/origin templated/redirect;
```
r-n-o, line 23: no need to create templated folders :)

kustomize/base/resources.yaml
```diff
@@ -20,10 +20,12 @@ spec:
...
                 envsubst '${TURNKEY_SIGNER_ENVIRONMENT}' < import/index.template.html > templated/import/index.html;
                 envsubst '${TURNKEY_SIGNER_ENVIRONMENT}' < import/standalone.template.html > templated/import/standalone.html;
                 envsubst '${TURNKEY_SIGNER_ENVIRONMENT}' < export/index.template.html > templated/export/index.html;
+                envsubst '${TURNKEY_SIGNER_ENVIRONMENT}' < oauth-origin/index.template.html > templated/oauth-origin/index.html;
+                envsubst '${TURNKEY_SIGNER_ENVIRONMENT}' < oauth-redirect/index.template.html > templated/oauth-redirect/index.html;
```
r-n-o, line 28: And no need for `envsubst` either

nginx.conf
moeodeh3: gotcha - thanks for the context

r-n-o (approved):

PR #114, opened 2026-02-06
Webpackified the export iframe!

ethankonk:


export/src/index.template.html
```diff
@@ -0,0 +1,19 @@
+<!--
+  This page is intended to be embedded in an iframe. When updating any functions in
+  index.html, please make the same updates in standalone.html.
+-->
+<!DOCTYPE html>
+<html class="no-js" lang="">
+  <head>
+    <link rel="icon" type="image/svg+xml" href="./favicon.svg" />
```
leeland-turnkey, line 8: We are going to need something like this or else we will get an error in preprod and production:
https://github.com/tkhq/frames/blob/main/export-and-sign/src/index.template.html#L8

 `verifyEnclaveSignature` will always throw in production

leeland-turnkey:
I believe we will need to update the `resources.yaml` file. The `export` folder runs `envsubst` against `export/index.template.html`. We will need to import changes from the `templated/export` tolder to the webpack dist of `export/dist`. Also the init container block needs to mirror the export-and-sign and import pattern: 
https://github.com/tkhq/frames/blob/main/kustomize/base/resources.yaml#L29-L56

export/webpack.config.js
```diff
@@ -0,0 +1,151 @@
...
+      modules: [path.resolve(__dirname, "node_modules"), "node_modules"],
+      // Don't use package.json from shared folder for module resolution
+      descriptionFiles: ["package.json"],
+      // Force resolution to start from context (frame directory) not file location
+      symlinks: false,
+    },
+    resolveLoader: {
+      modules: [path.resolve(__dirname, "node_modules"), "node_modules"],
+    },
+    optimization: {
```
leeland-turnkey, line 129: We will need to add deterministic in the optimization as well.

https://github.com/tkhq/frames/commit/04bc323682229c6b82b14e2b94cec68af2265396


```suggestion
    optimization: {
     // Reproducible builds so CI "dist matches committed" check passes
      moduleIds: "deterministic",
      chunkIds: "deterministic",
```

export/package.json
```diff
@@ -19,12 +22,27 @@
...
+    "webpack": "5.102.1",
+    "webpack-cli": "5.1.4",
+    "webpack-dev-server": "5.2.2",
+    "copy-webpack-plugin": "11.0.0",
+    "html-webpack-plugin": "5.5.3",
+    "css-loader": "6.8.1"
+  },
+  "dependencies": {
+    "@hpke/core": "^1.7.5",
+    "@noble/ed25519": "^3.0.0",
```
leeland-turnkey, line 44: Not sure if this matters, but export is running version `3.0.0` for this package while export-and-sign is running 2.0.0: https://github.com/tkhq/frames/blob/main/export-and-sign/package.json#L22

Another NIT:
We have carats as part of these dependencies while export-and-sign have the packages locked.

leeland-turnkey:
I would make sure the github actions for export-and-sign match export and re-run the CI build after the change. I believe there will be more errors and things we missed that will be uncovered: .[github/workflows/ci.yml](https://github.com/tkhq/frames/blob/main/.github/workflows/ci.yml)

leeland-turnkey (changes requested):

export/src/index.js
```diff
@@ -0,0 +1,320 @@
+// Vendor @hpke/core from https://esm.sh/@hpke/core@1.2.7
```
leeland-turnkey, line 1: We can remove this comment since it's managed by webpack

export/src/index.js
```diff
@@ -0,0 +1,320 @@
...
+  };
+
+  // Create a new div with the key material and append the new div to the body
+  const keyDiv = document.getElementById("key-div");
+  keyDiv.innerText = key;
+  for (let styleKey in style) {
+    keyDiv.style[styleKey] = style[styleKey];
+  }
+  document.body.appendChild(keyDiv);
+  TKHQ.applySettings(TKHQ.getSettings());
```
leeland-turnkey, line 159: we call `applySettings` here, but we don't in standalone.js. Do you know the reason why? I'm not sure if it is testing something or not: https://github.com/tkhq/frames/pull/114/changes#diff-1c20a0aebf2920f0209e01ecd1b5ef3cdf3739d0209015550ef3f0a18b83ba9cR156

leeland-turnkey (changes requested):

PR #115, opened 2026-02-12
added encrypted bundle caching + escrow key injection

ethankonk:
Added the ability to inject a P256 key that replaces the iframe's embedded key allowing for exported bundle caching.

#### 1. Dedicated escrow private key is created with Turnkey

```
await httpClient?.createPrivateKeys({
  privateKeys: [
     {
      privateKeyName: "<PK_NAME>",
      curve: "CURVE_P256", // MUST BE A P256 KEY
      privateKeyTags: [], // optional
      addressFormats: [] // not needed
     }
  ]
})
```

#### 2. Inject the escrow key into the iframe to replace the default embedded public key

> Nab the iframe's embedded public key to encrypt the escrow bundle too:
```
const targetPublicKey = await iframeClient.getEmbeddedPublicKey();
if (!targetPublicKey) {
    throw new Error("Failed to retrieve target public key from iframe.");
}

const { exportBundle } =
    (await httpClient?.exportPrivateKey({
    privateKeyId: escrowPrivateKeyId,
    targetPublicKey,
    })) || {};
```

> Inject the export bundle

```
iframeClient.replaceEmbeddedKey(organizationId, exportBundle);
```

 ```
 const { exportBundle } =
   (await httpClient?.exportWalletAccount({
     address: account.address,
     targetPublicKey: privateKey.publicKey, // escrow PK public key
   })) || {}; 
```

> Now we push the export bundle:
   
```
await iframeClient.injectKeyExportBundle(
    exportBundle,
    organizationId,
    KeyFormat.Solana, // or KeyFormat.Hexadecimal
    account.address // the address of the wallet account that we exported
);
```

---

Now all our encrypted bundles are decrypted in memory and ready to sign with! The escrow private key is never kept in persistent storage and must be re-injected whenever the iframe gets destroyed.

--- 

### Quick Demo 🎉 

https://www.loom.com/share/81c1893abf174a32beb8ae3d51086b6f

> The SDK side PR that includes this demo can be seen here: https://github.com/tkhq/sdk/pull/1200

Project doc: https://docs.google.com/document/d/16YXWrR75RnRmkM_gURLAzwrEG7FUzK_pF6Z-rQ5g0j8/edit?tab=t.0

### CODE REVIEW

File: export-and-sign/src/event-handlers.js [hunk 1]
```diff
@@ -324,6 +339,253 @@ async function onClearEmbeddedPrivateKey(requestId, address) {
...
+ * @param {string} organizationId - Organization ID to filter addresses by (only remove bundles matching the organization ID)
+ * @param {string|undefined} address - Optional wallet address
+ */
+function onClearStoredBundles(requestId, address, organizationId) {
+  if (address) {
+    TKHQ.removeEncryptedBundle(address, organizationId);
+    delete inMemoryKeys[address];
+  } else {
+    TKHQ.clearAllEncryptedBundles(organizationId);
+    inMemoryKeys = {};
```
leeland-turnkey: This logic removes all orgs right? I guess this use case wouldn't be hit because most users will only be a part of one org. There could be a chance in the future that there are multiple orgs within one company and this will blow away every org's in-memory key.

export-and-sign/src/event-handlers.js
```diff
@@ -173,9 +164,33 @@ async function onInjectKeyBundle(
       privateKey: key,
       format: keyFormat,
       expiry: new Date().getTime() + DEFAULT_TTL_MILLISECONDS,
-      keypair: cachedKeypair, // Cache the keypair for performance
+      keypair: cachedKeypair,
```
leeland-turnkey, line 167: How do we clear by address in my first comment? Since there is no `organizationId` field, is there a way to do org-aware clearing of in-memory keys?

export-and-sign/src/event-handlers.js
```diff
@@ -324,6 +339,253 @@ async function onClearEmbeddedPrivateKey(requestId, address) {
...
+
+        const encappedKeyBuf = TKHQ.uint8arrayFromHexString(
+          bundleData.encappedPublic
+        );
+        const ciphertextBuf = TKHQ.uint8arrayFromHexString(
+          bundleData.ciphertext
+        );
+
+        // HPKE-decrypt using the key
+        const keyBytes = await HpkeDecrypt({
```
leeland-turnkey: We have two variables named `keyBytes`: https://github.com/tkhq/frames/pull/115/changes#diff-24697156a650f90b0b6154cd01ada56112b8581e71d5c8122705f691a4772943R480

This inner `keyBytes` variable can work because of block scoping, but it could possibly cause confusion as well. I wonder if a rename would be good here?

export-and-sign/src/event-handlers.js
```diff
@@ -6,12 +6,79 @@ import * as nobleHashes from "@noble/hashes/sha512";
...
+    throw new Error("method not loaded");
+  }
+  const verified = await TKHQ.verifyEnclaveSignature(
+    bundleObj.enclaveQuorumPublic,
+    bundleObj.dataSignature,
+    bundleObj.data
+  );
+  if (!verified) {
+    throw new Error(
+      `failed to verify enclave signature: ${bundleObj.dataSignature}`
```
leeland-turnkey: I like this change over the old way we did it: https://github.com/tkhq/frames/pull/115/changes#diff-24697156a650f90b0b6154cd01ada56112b8581e71d5c8122705f691a4772943L53

We were exposing the entire bundle which probably was not necessary.

Nice work.

leeland-turnkey (changes requested):
Left some comments, but overall this is looking great!

export-and-sign/src/event-handlers.js
ethankonk: This is just clearing in-memory keys, not in local storage. There should only be your current org's keys in memory during each "session" so its fine to wipe those completely 😁

export-and-sign/src/event-handlers.js
ethankonk: `onClearStoredBundles` has an organizationId you must pass in, if the bundle you are trying to delete is not in your org, it will not allow you to delete it

See here:
```
/**
 * Removes a single encrypted bundle by address.
 * Only removes the bundle if it belongs to the specified organization.
 * @param {string} address - The wallet address to remove
 * @param {string} organizationId - Only remove if the bundle belongs to this org
 */
function removeEncryptedBundle(address, organizationId) {
  const bundles = getEncryptedBundles();
  if (!bundles || !bundles[address]) return;

  // Only remove if the bundle belongs to the specified organization
  if (bundles[address].organizationId !== organizationId) return;

  delete bundles[address];
  if (Object.keys(bundles).length === 0) {
    window.localStorage.removeItem(TURNKEY_ENCRYPTED_BUNDLES);
  } else {
    window.localStorage.setItem(
      TURNKEY_ENCRYPTED_BUNDLES,
      JSON.stringify(bundles)
    );
  }
}
```

export-and-sign/src/event-handlers.js
ethankonk: Copilot pointed that out for me 😅

leeland-turnkey: Copilot 🙏

leeland-turnkey (dismissed):
Everything looks good and the rest of my comments are NIT. LGTM

r-n-o:
This doesn't make a whole lot of sense to me: instead of exporting the wallet directly encrypted to the iframe, we export a private key to which the wallet is encrypted, and persist the wallet.

The result is similar: an export activity has to be performed in both cases, for signing (export private key vs. export wallet). What's the advantage?

(generally, the simpler the crypto the better, so this extra layer of persistence/wrapping makes me nervous)

ethankonk:
@r-n-o the issue this change is supposed to fix is allowing Pumpfun to persist a bunch of wallets without having to export them all each session (they need to export 30 to 100+ wallets each session). For normal client side signature use you definitely should just use the normal flow (what you described) but in this case, they want tens of export bundles stored in the frames persistent storage for as long as possible

So like you mentioned, to initiate signing, instead of having to export and load a 100 wallets (which takes up to 2 mins), you just need to inject the exported key that can decrypt all the bundles stored in the iframe local storage

ethankonk:
You can find their exact use case in the project doc linked here

https://docs.google.com/document/d/16YXWrR75RnRmkM_gURLAzwrEG7FUzK_pF6Z-rQ5g0j8/edit?tab=t.0

ethankonk:
The issue this change is supposed to fix is allowing an important customer to persist a bunch of wallets without having to export them all each session (they need to export many wallets each session). For normal client side signature use you definitely should just use the normal flow (what you described) but in this case, they want tens of export bundles stored in the frames persistent storage for as long as possible

So like you mentioned, to initiate signing, instead of having to export and load a 100 wallets (which takes up to 2 mins), you just need to inject the exported key that can decrypt all the bundles stored in the iframe local storage

leeland-turnkey:
Code wise, this looks good to me. I know @r-n-o had some thoughts, so I'll wait for his final approval.

export-and-sign/src/event-handlers.js
```diff
@@ -6,12 +6,79 @@ import * as nobleHashes from "@noble/hashes/sha512";
 // Persist keys in memory via mapping of { address --> pk }
 let inMemoryKeys = {};
 
+// Injected decryption key -- held in memory only, never persisted.
+// When set, decryptBundle uses this P-256 JWK instead of the iframe's embedded key.
+let decryptionKey = null;
```
r-n-o: maybe name this "injectedEmbeddedKey" to avoid confusion? Basically we're offering an override here.

(or even `embeddedKeyOverride`?)

export-and-sign/src/event-handlers.js
```diff
@@ -6,12 +6,79 @@ import * as nobleHashes from "@noble/hashes/sha512";
...
+  }
+  if (!bundleObj.dataSignature) {
+    throw new Error('missing "dataSignature" in bundle');
+  }
+  if (!bundleObj.enclaveQuorumPublic) {
+    throw new Error('missing "enclaveQuorumPublic" in bundle');
+  }
+
+  if (!TKHQ.verifyEnclaveSignature) {
+    throw new Error("method not loaded");
```
r-n-o: "method TKHQ.verifyEnclaveSignature not loaded"?

export-and-sign/src/event-handlers.js
```diff
@@ -6,12 +6,79 @@ import * as nobleHashes from "@noble/hashes/sha512";
...
+  }
+  const verified = await TKHQ.verifyEnclaveSignature(
+    bundleObj.enclaveQuorumPublic,
+    bundleObj.dataSignature,
+    bundleObj.data
+  );
+  if (!verified) {
+    throw new Error(
+      `failed to verify enclave signature: ${bundleObj.dataSignature}`
```
r-n-o: If we want to log to be useful we should log the public key, the message, and the signature. Either of these could be wrong and lead to a signature verification failure!

export-and-sign/src/event-handlers.js
```diff
@@ -173,9 +164,33 @@ async function onInjectKeyBundle(
...
+/**
+ * Function triggered when INJECT_KEY_EXPORT_BUNDLE event is received.
+ * @param {string} requestId
+ * @param {string} organizationId
+ * @param {string} bundle
+ * @param {string} keyFormat
+ * @param {string} address
+ * @param {Function} HpkeDecrypt // TODO: import this directly (instead of passing around)
+ */
+async function onInjectKeyBundle(
```
r-n-o: Let's name the event `OVERRIDE_EMBEDDED_KEY` and this function `onOverrideEmbeddedKey` for clarity. Otherwise there's potential confusion between injecting a wallet/key bundle and this functionality

export-and-sign/src/event-handlers.js
```diff
@@ -324,6 +339,44 @@ async function onClearEmbeddedPrivateKey(requestId, address) {
...
+  // The decrypted payload is a raw 32-byte P-256 private key scalar.
+  const keyBytes = await decryptBundle(bundle, organizationId, HpkeDecrypt);
+
+  // Convert raw P-256 bytes to a full JWK (derives public key via WebCrypto)
+  const keyJwk = await rawP256PrivateKeyToJwk(new Uint8Array(keyBytes));
+
+  // Store in module-level variable (memory only)
+  decryptionKey = keyJwk;
+
+  TKHQ.sendMessageUp("DECRYPTION_KEY_INJECTED", true, requestId);
```
r-n-o: let's name this "EMBEDDED_KEY_REPLACED" (or `EMBEDDED_KEY_OVERRIDE_SET` or `EMBEDDED_KEY_OVERRIDDEN` if you choose to take my naming suggestion for the event/function) for consistency?

export-and-sign/src/event-handlers.js
```diff
@@ -324,6 +339,44 @@ async function onClearEmbeddedPrivateKey(requestId, address) {
...
+/**
+ * Handler for REPLACE_EMBEDDED_KEY events.
+ * Decrypts a P-256 private key bundle using the iframe's embedded key and
+ * replaces the embedded key with it for subsequent bundle decryptions.
+ * @param {string} requestId
+ * @param {string} organizationId
+ * @param {string} bundle - v1.0.0 bundle containing the P-256 private key
+ * @param {Function} HpkeDecrypt
+ */
+async function onReplaceEmbeddedKey(
```
r-n-o: Nit: this isn't really replacing, it's setting an override (because you can undo it later!). So "OVERRIDE_EMBEDDED_KEY" or "SET_EMBEDDED_KEY_OVERRIDE" , something like that

r-n-o (commented):
overall looking great! Some minor naming-related suggestions below

export-and-sign/src/event-handlers.js
ethankonk: Going to omit the message in this case since that would be the encrypted bundle, but added in the enclave quorum pubkey!

export-and-sign/src/event-handlers.js
ethankonk: Missed this one!

export-and-sign/src/event-handlers.js
```diff
@@ -6,12 +6,79 @@ import * as nobleHashes from "@noble/hashes/sha512";
 // Persist keys in memory via mapping of { address --> pk }
 let inMemoryKeys = {};
 
+// Injected decryption key -- held in memory only, never persisted.
+// When set, decryptBundle uses this P-256 JWK instead of the iframe's embedded key.
+let injectedDecryptionKey = null;
```
r-n-o: `injectedEmbeddedKey` or `embeddedKeyOverride` (slight preference for `embeddedKeyOverride` because "injected" and "embedded" in the same name sort of clash!)

ethankonk: Oh right, brain farted when I made that change 🤦

export-and-sign/index.test.js
```diff
@@ -852,3 +854,307 @@ describe("Event Handler Expiration Flow", () => {
     }
   });
 });
+
+describe("Decryption Key Override", () => {
```
r-n-o: Embedded Key Override

r-n-o (dismissed):

r-n-o (approved):

PR #117, opened 2026-03-04
Address audit findings in export-and-sign/import

ethankonk:
- Constant-time string comparison (timingSafeEqual) for org ID and enclave quorum public key checks to prevent timing side-channel attacks
- In-place key map mutation instead of spread-copy to avoid multiplying key material references on the V8 heap
- Zero sensitive Uint8Array buffers (secretKey, privateKeyBytes, pkcs8) on all code paths including error paths via finally blocks
- Origin validation on incoming postMessage events using allowedOrigin captured during the TURNKEY_INIT_MESSAGE_CHANNEL handshake
- Use captured parent origin as targetOrigin in sendMessageUp instead of wildcard '*' to prevent message eavesdropping
- Reduce embedded key localStorage TTL from 48h to 4h to limit exposure window if storage is compromised
- Add tests for all new security behaviors: timing-safe comparison, parent origin capture/send, TTL enforcement, key clearing with buffer zeroing, and origin-scoped postMessage


ethankonk:
### Code review

Found 1 issue:

1. **Origin validation bypass via empty-string `event.origin`**: The origin check `if (allowedOrigin && event.origin && event.origin !== allowedOrigin)` requires `event.origin` to be truthy. In browsers, `event.origin` is `""` (empty string, falsy) for messages from sandboxed iframes (`<iframe sandbox>`), `file://` pages, `data:` URIs, and `blob:` URLs. This means messages from null-origin contexts bypass the validation entirely, even after `allowedOrigin` has been captured during the handshake. The fix is to check `allowedOrigin !== null` instead of relying on truthiness: `if (allowedOrigin !== null && event.origin !== allowedOrigin)`.

https://github.com/tkhq/frames/blob/1a8dac4e4f9d59dcf10b283b799285bce02f3b68/export-and-sign/src/event-handlers.js#L739-L745

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

export-and-sign/src/turnkey-core.js
```diff
@@ -3,6 +3,32 @@ import * as nobleHashes from "@noble/hashes/sha512";
...
+ */
+function timingSafeEqual(a, b) {
+  const enc = new TextEncoder();
+  const aBuf = enc.encode(a);
+  const bBuf = enc.encode(b);
+  if (aBuf.length !== bBuf.length) {
+    let diff = 1;
+    const len = Math.min(aBuf.length, bBuf.length);
+    for (let i = 0; i < len; i++) {
+      diff |= aBuf[i] ^ bBuf[i];
```
andrewkmin, line 21: This one is a valid finding -- we don't do anything with the `diff` that we mutate

export-and-sign/src/event-handlers.js
```diff
@@ -3,19 +3,93 @@ import { Keypair, VersionedTransaction } from "@solana/web3.js";
...
+  // Encode to bytes so we XOR fixed-width units (UTF-8 bytes)
+  const aBuf = textEncoder.encode(a);
+  const bBuf = textEncoder.encode(b);
+  if (aBuf.length !== bBuf.length) {
+    // Length mismatch already leaks info, but we still do constant-time
+    // comparison over the shorter length to avoid additional leakage.
+    let diff = 1; // already know they differ
+    const len = Math.min(aBuf.length, bBuf.length);
+    for (let i = 0; i < len; i++) {
+      diff |= aBuf[i] ^ bBuf[i];
```
andrewkmin, line 63: same here -- we should also DRY this up (write once, then import)

export-and-sign/src/event-handlers.js
```diff
@@ -614,6 +741,18 @@ function addDOMEventListeners() {
  */
 function initMessageEventListener(HpkeDecrypt) {
   return async function messageEventListener(event) {
+    // SECURITY: Validate event.origin against the allowlist captured at init time.
```
andrewkmin, line 744: let's remove this from the scope of this change for now

export-and-sign/src/event-handlers.js
```diff
@@ -769,6 +908,12 @@ export function initEventHandlers(HpkeDecrypt) {
         event.data["type"] == "TURNKEY_INIT_MESSAGE_CHANNEL" &&
         event.ports?.[0]
       ) {
+        // SECURITY: Capture the parent origin from the init handshake.
```
andrewkmin, line 911: same with this. I think we can ticket it / consider tackling in a separate workstream

export-and-sign/src/turnkey-core.js
```diff
@@ -84,7 +112,14 @@ async function verifyEnclaveSignature(
     );
   }
 
-  if (enclaveQuorumPublic !== TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY) {
+  // SECURITY: Use constant-time comparison to prevent timing side-channel
+  // attacks that could leak the enclave quorum public key byte-by-byte.
+  if (
```
andrewkmin, line 117: this is public information; we can remove the `timingSafeEqual`  check

shared/turnkey-core.js
```diff
@@ -404,7 +468,14 @@ async function verifyEnclaveSignature(
 
   // todo(olivia): throw error if enclave quorum public is null once server changes are deployed
   if (enclaveQuorumPublic) {
-    if (enclaveQuorumPublic !== TURNKEY_SIGNER_ENCLAVE_QUORUM_PUBLIC_KEY) {
+    // SECURITY: Use constant-time comparison to prevent timing side-channel
```
andrewkmin, line 471: can remove this as well

shared/turnkey-core.js
andrewkmin, line 911: hm this is a good find, wonder how it slipped in. for now, if `element` is indeed unused, we can remove

shared/turnkey-core.js
```diff
@@ -5,15 +5,61 @@ import { bech32 } from "bech32";
...
+ * TTL for the embedded ECDH private key stored in localStorage.
+ *
+ * SECURITY NOTE: The P-256 ECDH private key is stored as a JSON-serialized JWK
+ * in localStorage, where it cannot be reliably zeroed (localStorage values are
+ * immutable strings managed by the browser). The ideal solution would be to store
+ * a non-extractable CryptoKey in IndexedDB, but that is a larger refactor.
+ * As a mitigation, we keep the TTL as short as practical (4 hours) to limit the
+ * window of exposure if the storage is compromised.
+ */
+const TURNKEY_EMBEDDED_KEY_TTL_IN_MILLIS = 1000 * 60 * 60 * 4;
```
andrewkmin, line 52: I think we can keep parts of this comment, but save other parts (like `The ideal solution would be to store a non-extractable CryptoKey in IndexedDB, but that is a larger refactor.`) for a separate ticket.

let's undo these changes for now, as I think going down to 4 hours will be too aggressive

shared/turnkey-core.js
```diff
@@ -5,15 +5,61 @@ import { bech32 } from "bech32";
...
+  if (typeof a !== "string" || typeof b !== "string") return false;
+  const enc = new TextEncoder();
+  const aBuf = enc.encode(a);
+  const bBuf = enc.encode(b);
+  if (aBuf.length !== bBuf.length) {
+    // Length mismatch already leaks info; still iterate to avoid further leakage.
+    let diff = 1;
+    const len = Math.min(aBuf.length, bBuf.length);
+    for (let i = 0; i < len; i++) {
+      diff |= aBuf[i] ^ bBuf[i];
```
andrewkmin, line 27: same comment as in the other iframe: let's revisit this logic

shared/turnkey-core.js
```diff
@@ -5,15 +5,61 @@ import { bech32 } from "bech32";
...
+/**
+ * Constant-time string comparison to prevent timing side-channel attacks.
+ * Standard `===` / `!==` short-circuits on the first differing byte, leaking
+ * information about how many leading bytes match. This XOR-based approach
+ * always compares every byte regardless of mismatches.
+ * @param {string} a
+ * @param {string} b
+ * @returns {boolean} true if strings are equal
+ */
+function timingSafeEqual(a, b) {
```
andrewkmin, line 17: actually on second thought, we can fix this implementation and eventually throw it into https://github.com/tkhq/frames/tree/main/shared

import/src/index.js
```diff
@@ -123,6 +123,10 @@ window.addEventListener(
       // remove the message event listener that was added in the DOMContentLoaded event
       messageListenerController.abort();
 
+      // Capture the parent origin for use as targetOrigin in sendMessageUp,
```
andrewkmin, line 126: want to roll this out separately ideally (just to reduce blast radius)

export-and-sign/src/event-handlers.js
```diff
@@ -354,17 +456,33 @@ async function onSetEmbeddedKeyOverride(
...
-  injectedEmbeddedKey = keyJwk;
-
-  TKHQ.sendMessageUp("EMBEDDED_KEY_OVERRIDE_SET", true, requestId);
+  let keyBytes = null;
+  try {
+    // Decrypt the private key using the iframe's embedded key.
+    // The decrypted payload is a raw 32-byte P-256 private key scalar.
+    const decrypted = await decryptBundle(bundle, organizationId, HpkeDecrypt);
+    keyBytes =
+      decrypted instanceof Uint8Array ? decrypted : new Uint8Array(decrypted);
```
andrewkmin, line 465: ah, this accentuates the need for type safety :')

export-and-sign/src/event-handlers.js
```diff
@@ -135,14 +211,21 @@ async function loadKeyIntoMemory(address, keyBytes, keyFormat, organizationId) {
...
+        privateKeyBytes.subarray(0, 32)
+      );
+      const publicKeyBytes = TKHQ.getEd25519PublicKey(privateKeyHex);
+      key = await TKHQ.encodeKey(privateKeyBytes, keyFormat, publicKeyBytes);
+    } else {
+      key = await TKHQ.encodeKey(privateKeyBytes, keyFormat);
+    }
+  } finally {
+    // SECURITY: Zero the raw private key bytes immediately after encoding.
+    // The encoded `key` is a JS string (hex or base58) which we cannot zero,
```
andrewkmin, line 226: maybe we shouuuld rearchitect such that no private key material is ever a JS string? will have to think about this one; ultimately if it's displayed as a string in an iframe (which it sort of needs to be in order to be a functional export iframe), then we don't gain much from this

andrewkmin (commented):
a couple gaps in tkintern's implementation 😛 

main takeaways:
- reduce the scope of this to exclude origin checks (perhaps in a separate PR)
- on constant time helper function: let's DRY it up, fix/test the implementation, and move it to `shared`
- we could explore creating a new class for the inMemoryKeys object and override the `delete` method (instead of calling `zeroKeyEntry()`). don't feel too strongly here

andrewkmin (changes requested):

shared/turnkey-core.js
```diff
@@ -5,11 +5,41 @@ import { bech32 } from "bech32";
...
```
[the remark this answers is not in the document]
ethankonk: @andrewkmin we didn't wanna do this in this PR right?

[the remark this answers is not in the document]
andrewkmin: mmm yeah let's leave it out

mssabr01:
We've addressed this in https://github.com/tkhq/frames/pull/129

PR #122, opened 2026-07-06
Add EVM support to the export-and-sign signing flow

t-vila:
Extends the export-and-sign iframe to sign EVM transactions and messages with an injected private key, alongside the existing Solana support. Signing happens entirely in-memory inside the iframe; key material never reaches the parent page.

- onSignTransaction — new ETHEREUM branch. Takes a 0x-prefixed serialized unsigned transaction, signs it, and returns the broadcast-ready serialized signed transaction (via viem parseTransaction → account.signTransaction).

- onSignMessage — new ETHEREUM branch. EIP-191 personal_sign over the UTF-8 message, matching how the Solana path treats messages.

- New helpers getOrCreateEthereumAccount (lazy + cached, mirrors getOrCreateKeypair) and signEthereumTransaction.

- Adds viem@2.45.0.

**Design note — contract is symmetric with Solana**

The EVM path deliberately mirrors the Solana contract: the caller builds and serializes the unsigned transaction with whatever library it likes (ethers/viem/web3/…) and hands the iframe serialized hex; the iframe returns serialized signed hex. This keeps the iframe build-agnostic rather than coupling callers to a specific transaction-object shape.

```
// EVM
signTransaction(
  { transaction: serializedUnsignedTxHex, type: TransactionType.Ethereum },
  address,
);
// Solana (unchanged)
signTransaction(
  { transaction: serializedTxHex, type: TransactionType.Solana },
  address,
);
```

A JSON tx-object input form can be added later as a superset (branch on 0x vs {) if we want the ergonomic alternative — not needed now.

No changes to `@turnkey/iframe-stamper` required: `TransactionType.Ethereum / MessageType.Ethereum` already exist. `loadKeyIntoMemory` is untouched — the viem account is derived lazily at sign time, so no `keyFormat → curve` coupling.

**Tests**

New `EVM Signing` suite in `index.test.js`

- signs an EVM transaction and asserts the result recovers to the injected key's address
- signs an EVM message (EIP-191) and asserts it recovers to the injected key's address
- rejects an unsupported transaction type

Full suite: 39/39 passing, lint clean.

Manual testing using the updated example in https://github.com/tkhq/sdk/pull/1432

Closes REQ-418


Dockerfile
andrewkmin, line 1: qq -- what prompted the changes to this file? I don't see any changes that would necessitate updating this dockerfile. If there are fixes to be made that aren't specific to this feature (enabling EVM support), then I'd recommend taking this out of the current PR and ticketing/tackling in a separate PR, just to decouple and reduce rollout risk

t-vila: It was actually required by this change, not sure if there's a better solution tho ?

Enabling EVM adds `viem`, and the docker job builds `linux/amd64,linux/arm64` via `buildx`, so the arm64 leg runs under QEMU emulation. 
The old Dockerfile built export-and-sign in-container (`RUN cd export-and-sign && npm ci && npm run build`), and under emulation that step dies pulling viem's dependency tree:

```
npm error network read ECONNRESET  (errno -104)
ERROR: failed to solve: process "/dev/.buildkit_qemu_emulator /bin/sh -c cd export-and-sign && npm ci && npm run build" did not complete successfully: exit code: 152
```

The fix switches export-and-sign to copy its committed dist/ — identical to how the export frame is already handled (COPY export …) — and integrity is enforced by the existing build-check (export-and-sign) job, which rebuilds and diffs against the committed dist/ on every run.

andrewkmin: I wonder if we could just have the original Dockerfile and then have

```
FROM --platform=$BUILDPLATFORM node:18-bullseye-slim AS builder
```
for compatibility

cc @janjakubnanista

janjakubnanista: Yeah lemme see what's up there

janjakubnanista: @t-vila take a look at https://github.com/tkhq/frames/pull/124 and lmk if that fixes your problem - there was a bit of an issue with the docker build context (it would only affect local builds though)

janjakubnanista (approved):

andrewkmin (approved):

PR #128, opened 2026-07-17
chore: Switch to turborepo

janjakubnanista:
### In this PR

- Switch from `cd xxx && npm i` to a `turbo` based setup that mirrors `sdk` (in fact, it has been largely if not completely copied from there). This brings a lot of goodies in
- Add TypeScript support (only affecting test files though)
- Fix some JSDOC comments to make TypeScript happy
- Fix JSDOM test setup so that we can remove the triplicated `verifyEnclaveSignature` function from tests. Basically the tests were testing that a mocked function works as expected


justinformentin:
We currently have doc outlining a plan to do just this - convert to a monorepo and typescript. 

We are going to approach this as a multi-step change to make it possible to review and confirm everything still works and behaves the exact same way. This is too large of a change for a single PR.

janjakubnanista:
@justinformentin same thoughts - although turbo is just one of many pieces of the puzzle.

https://github.com/tkhq/frames/pull/125 contains everything needed to get this repo up to speed:

- `turbo`
- Removing that nasty string replacement in k8s resources in favour of `nginx` SSI
- Optimized Dockerfile
- Dev setup
- E2E tests coming

The PR is massive and the idea is to split it up into 10+ I guess steps. But to make it safe, E2E regression test suite is essential. To make things difficult, of course is much harder to add it first, then introduce turbo, but it's doable.

Also for browser testing (`vitest` using `playwright`), a runtime code change is required. The test runtime runs as an `iframe`, we then create yet another `iframe` and our `iframe` now triggers an error in `isDoublyIframed`.

PR #130, opened 2026-08-07
fix(import): extend cross-origin MessageChannel validation to import frame (INT-783)

justinformentin:
**Summary**

Extends the cross-origin `MessageChannel` validation from INT-697 / [PR #129](https://github.com/tkhq/frames/pull/129) to the import frame.

PR #129 hardened the `export-and-sign` frame's `TURNKEY_INIT_MESSAGE_CHANNEL` handler against opaque origins and non-parent senders. The `import` frame had the same pre-existing gap and was not touched by PR #129.

**Linear:** INT-783 https://linear.app/turnkey/issue/INT-783

**Security ticket addressed:** ENG-4598 (TKA-20260806-014) https://linear.app/turnkey/issue/ENG-4598 — "The hosted import frame accepts the first `MessageChannel` from any sender." This PR closes the channel-seizure vector by requiring the sender to be the direct parent with a concrete origin.

> Note: ENG-4598 also references `onInjectImportBundle()` org/user binding when a caller omits those fields. That permissive-omission behavior is tracked separately in ENG-4597 and is **not** changed here — this PR only hardens the channel handshake gate.

**Before (import/src/index.js)**
```js
if (
  event.data &&
  event.ports?.[0]
) {
```

**After**
```js
if (
  event.data &&
  event.source === window.parent &&
  event.origin &&
  event.origin !== "null" &&
  event.ports?.length === 1
) {
```

**What changed**
- `event.source === window.parent` — only the direct parent frame may establish the channel
- `event.origin && event.origin !== "null"` — opaque/sandboxed origins are rejected
- `event.ports?.length === 1` — exactly one transferred port required (was: any truthy ports[0])

The import frame is intentionally simpler than export-and-sign (no persistent embedded key, no telemetry, no legacy origin binding) — only the gate condition is updated.

**Validation**
- ESLint: 0 errors, 4 pre-existing `no-console` warnings (none from this change)
- Prettier: all files pass
- Jest: 25/25 tests pass (19 pre-existing + 6 new gate tests covering: accept valid case, reject non-parent source, reject opaque/null origin, reject empty origin, reject 0 ports, reject 2 ports)
- Production build: compiled successfully, new bundle hash `45de86a592cd0d80ccd0`

**Reference**
- PR #129 (merged 2026-08-06): https://github.com/tkhq/frames/pull/129
- INT-697: https://linear.app/turnkey/issue/INT-697
- ENG-4598 (this PR): https://linear.app/turnkey/issue/ENG-4598


import/src/index.js
```diff
@@ -121,7 +121,10 @@ window.addEventListener(
     if (
       event.data &&
       event.data["type"] == "TURNKEY_INIT_MESSAGE_CHANNEL" &&
-      event.ports?.[0]
+      event.source === window.parent &&
```
fainashalts, line 167: this validates only the `MessageChannel` handshake. The legacy listener registered in `DOMContentLoaded` still attaches `messageEventListener` directly, so until a channel is established `INJECT_IMPORT_BUNDLE`, `EXTRACT_*` and `APPLY_SETTINGS` accept messages from any source or origin if I'm reading this right.

The new org/user checks don't close this path because those expected values come from the same untrusted message. A non-parent sender can provide a valid signed bundle with matching IDs and overwrite the persisted target key or trigger extraction/reset operatioons. 

Should mirror #129's legacy wrapper: require the direct parent, reject empty/opaque origins, bind subsequent traffic to the first accepted origin, and restrict responses to that origin. Should also add regression tests for operational messages from non-parent and opaque-origin senders.

PR #131, opened 2026-08-07
fix(auth): scope embedded key to parent origin + validate MessageChannel sender (ENG-4596)

justinformentin:
## Summary

This PR ports the INT-697 cross-origin hardening pattern to the **auth** frame, closing two related vulnerabilities (ENG-4596 / TKA-20260806-016, Medium).

The auth frame is structurally different from export-and-sign: the source is entirely inline in `auth/index.html` (no `src/` directory, no build step). This PR adapts the same security patterns faithfully to the inline-script architecture.

---

## Vuln A — MessageChannel gate accepts any sender

**Before:** The `TURNKEY_INIT_MESSAGE_CHANNEL` handler gated only on `event.ports?.[0]`, allowing any window (not just the direct parent) to establish the privileged MessageChannel and receive subsequent messages via the port.

```js
// Before
if (
  event.data &&
  event.ports?.[0]
)
```

**After:** The gate now validates source, origin, and port count — mirroring the INT-697 export-and-sign fix:

```js
// After
if (
  event.data &&
  event.source === window.parent &&
  event.origin &&
  event.origin !== "null" &&
  event.ports?.length === 1
)
```

---

## Vuln B — Parent-agnostic embedded key

**Before:** A single P-256 private key was persisted in localStorage under the fixed name `"TURNKEY_EMBEDDED_KEY"`, created at DOMContentLoaded before any parent origin was known. The same key (and thus the same `PUBLIC_KEY_READY` public key) was served to every embedder. An attacker could replay an auth bundle encrypted to that key from any origin.

**After:** The inline TKHQ IIFE now tracks an `embeddedKeyState` object with two modes:

- **`persistent`** — key stored in localStorage under `TURNKEY_EMBEDDED_KEY_V2:<encodeURIComponent(origin)>`. Each parent origin gets its own isolated key slot.
- **`ephemeral`** — key stored in memory only (for legacy `@turnkey/iframe-stamper < 2.1.0` clients). Unique to this document; never persisted.

**Key new helpers added inline:**
- `validateParentOrigin(origin)` — validates non-empty, non-opaque, valid serialized origin
- `purgeLegacyEmbeddedKey()` — removes old `TURNKEY_EMBEDDED_KEY` (never migrated)
- `initEmbeddedKey(origin)` — now REQUIRES an origin; creates/reuses V2 scoped persistent key
- `initEphemeralEmbeddedKey()` — creates in-memory-only key; no-ops if persistent key is active
- `getBoundOrigin()` — returns the currently bound origin
- Updated `getEmbeddedKey()`, `setEmbeddedKey()`, `resetEmbeddedKey()` to dispatch based on state mode

**Flow changes:**
- Standalone mode (DOMContentLoaded): `initEmbeddedKey(window.location.origin)`
- Embedded mode (DOMContentLoaded): `initEphemeralEmbeddedKey()` for legacy clients
- MessageChannel handshake: `initEmbeddedKey(event.origin)` supersedes ephemeral; rolls back on failure
- Legacy embedded path: binds to first sender origin; rejects different origins

---

## Tests

Added/updated in `auth/index.test.js` (34 total, all passing):

**New Vuln B tests:** different origins produce different keys, refuses second origin binding, rejects invalid origins, purges legacy key, ephemeral key stays in memory only, getBoundOrigin coverage.

**New Vuln A tests (channel gate):** rejects non-parent source, empty origin, opaque `"null"` origin, 0 ports, 2 ports; accepts valid event.

---

## Validation Results

```
prettier:check  — All matched files use Prettier code style!
eslint          — No issues found
jest            — 34 passed, 34 total (5.2s)
```

No build step exists for the auth frame (inline HTML is the source). No dist artifacts to commit.

---

## References

- Linear: ENG-4596 — https://linear.app/turnkey/issue/ENG-4596
- Sibling fix for export-and-sign: INT-697 / PR #129
- Sibling fix for import: PR #130


auth/index.html
```diff
@@ -158,29 +158,143 @@ <h2>Message log</h2>
...
+          }
+
+          purgeLegacyEmbeddedKey();
+
+          var storageKey =
+            ORIGIN_SCOPED_EMBEDDED_KEY_PREFIX +
+            ":" +
+            encodeURIComponent(validatedOrigin);
+
+          embeddedKeyState = {
```
fainashalts: This activates persistent mode before key generation and storage complete. `initEphemeralEmbeddedKey()` then returns immediately when it sees `mode === "persistent"`, so concurrent bootstrap can observe `getEmbeddedKey() === null`. 

If `localStorage.setItem()` throws, the previous ephemeral key is also lost and the claimed legacy fallback no longer works.

We should retain the previous state, associate a readiness promise with the new persistent state, make concurrent callers await it, and restore the previous state on rejection, as in #129. Should also add coverage for the initialization race and storage-failure rollback.

auth/index.html
```diff
@@ -1163,27 +1320,87 @@ <h2>Message log</h2>
...
+            // Only the direct parent, with a real (non-opaque) origin, may drive
+            // the legacy path.
+            if (event.source !== window.parent) {
+              return;
+            }
+            if (!event.origin || event.origin === "null") {
+              return;
+            }
+            if (legacyParentOrigin === null) {
+              legacyParentOrigin = event.origin;
```
fainashalts, line 1463: This binds inbound legacy messages to the first origin, but outbound legacy responses are still sent by `sendMessageUp()` using `postMessage(..., "*")`. If the parent navigates while an asynchronous request is running, the response can be delivered to the replacement origin.

After the first valid legacy request, we should store this validated origin as the outbound `targetOrigin` and use it for subsequent responses. Only the initial `PUBLIC_KEY_READY`, sent before an origin is known, should use "*".

auth/index.html
```diff
@@ -1207,42 +1511,39 @@ <h2>Message log</h2>
...
-              event.data.dangerouslyOverrideIframeKeyTtl &&
-              typeof event.data.dangerouslyOverrideIframeKeyTtl === "number" &&
-              event.data.dangerouslyOverrideIframeKeyTtl > 0
-            ) {
-              iframeKeyTtl = event.data.dangerouslyOverrideIframeKeyTtl;
+            var targetPubHex;
+            try {
+              // Vuln B fix: Supersede the ephemeral key (if any) with a persistent
+              // key scoped to the browser-authenticated parent origin (INT-697).
+              await TKHQ.initEmbeddedKey(event.origin);
```
fainashalts: This removes support for `event.data.dangerouslyOverrideIframeKeyTtl`. Before this PR, a positive numeric override was passed to `initEmbeddedKey` but now every persistent key is stored with the fixed 48-hour TTL. Callers requesting a shorter lifetime will silently retain the decryption key much longer than configured.

We should preserve the validated override as a second argument, e.g. `initEmbeddedKey(event.origin, iframeKeyTtl)`, and use it in `setItemWithExpiry`. Let's also add a test confirming the handshake override controls the stored expiry. 

Was removing this intentional? If so we should discuss it explicitly as an API/security decision.

